PR #8, opened 2021-03-22
Exclude facets + consolidation with restoration api

Semtexcz:


oarepo_ui/facets.py
```diff
@@ -1,7 +1,38 @@
...
+        ret['terms']['missing'] = missing
+    return ret
+
+
+# ROLE FACETS
+class RoleFacets:
+    def __init__(self, **kwargs):
+        self.data = kwargs
 
+    def current_role(self):
```
mirekys, line 26: Co ostatni role uzivatele krom anonymous/authenticated

mesemus: imho se to planovalo tak, ze ostatni role se resi v podedene classe - protoze clovek muze byt ve vic rolich ale z hlediska facet pro tohle musis vybrat jednu ....

Za me ok, mergnete + release? Dik

PR #56, opened 2023-07-25
Implement multilingual text input

mirekys:


Ducica:
@mirekys Ahoj Miro, I pushed my changes to the multilingual component. Please take a look. We can discuss tomorrow. Thanks!

Ducica:
@mirekys Ahoj Miro, I think there is a small issue with obj2array func in https://github.com/oarepo/oarepo-ui/pull/56/files#diff-a1e07c1856e416e328185387667e621add86c9864336a31c929ae6a0d0eac69b

`export const object2array = (obj, keyName, valueName) => {
  // Transforms object to array of objects.
  // Each key of original object will be stored as value of `keyName` key.
  // Each value of original object will be stored as value of `valueName` key.

  _map(obj, (value, key) => ({
    [keyName]: key,
    [valueName]: value,
  }));
};`

The function does not return anything.

Ducica:
@mirekys Maybe not the best place to post this comment, but it is also related to these multilingual issues. It seems that the type i18n

image

Is literally used only in this one place for additional titles i.e. I am not sure if it warrants a separate component or should I just code it into the additionalTitles component?

mirekys:
I just implemented that input in https://github.com/oarepo/oarepo-ui/pull/65, you can now use that in the MultilingualTextInput by wrapping it in an aarrayfield

Ducica:
@mirekys Ciao Mirku, I made the following changes:

- implemented i18nRichInput component
- use both I18n inputs in multilingualTextInput

Ducica:
@mirekys Ahoj, I think there are two small issues:

1. Regarding the width of the rich input field, yes if we set it to 16 it looks ok in the begining, however if you paste a block of text inside of the input, the layout somehow changes and now the remove button protrudes to the right side:
image and I honestly don't understand how this is even possible. Maybe some richfield (CKeditor) particularity?
2. They key should actually be available options as below. It is because if not, then only that particular language select rerenders and others don't when you change it, which makes it possible sometimes to choose the same language twice, or sometimes, after you clear one language from adjacent language select, it is not available in the other select input immediatelly
image

mirekys:
> 2. They key should actually be available options as below. It is because if not, then only that particular language select rerenders and others don't when you change it, which makes it possible sometimes to choose the same language twice, or sometimes, after you clear one language from adjacent language select, it is not available in the other select input immediatelly
>    image

The correct way instead of relying on huge component keys (list of ~158 items for each input :) ) to do this would be to useState for available languages on multilingual input, and update state whenever any language input changes, which then rerenders the children using that state. But I haven't been able to do this myself because of ArrayField puts difficulty into that

PR #65, opened 2023-08-11
[mirekys/fe 42] implement i18ntextinput component

mirekys:


Ducica:
@mirekys Ciao Miro, I notice in relatedSelectField, you removed the part of the code that kept the state in Formik in appropriate format i.e. {id: value} or [{id:val}, {id, val}], does this mean that we no longer wish the components to manage and keep the formik's state in appropriate shape themselves, but we want to manipulate data before it goes to inputs and also before it is supposed to be submitted?

PR #68, opened 2023-08-17
fixed issue with onbeforesubmit

Ducica:
@mirekys Ahoj Miro, I fixed this issue by simply passing default onBeforeSubmit func that takes value and returns it. I tried it out and it works fine. It seems that we have different prettifier or something that is marking all lines in the codes as changed, so we should probably consider sitting down and making uniform settings between all of us who work on FE, to reduce the clutter and make PRs easier to check.


mirekys:
Hi @Ducica , thank a lot for fixing this 👍, regarding the code formatting, the mistake is probably on my side, because of:
`Editor: Format On Save: Off` vscode setting on my side. I've turned it on. I believe the current nrp tooling sets the same formatter for everyone.

mirekys (approved):

Ducica:
@mirekys Welcome!:)

PR #73, opened 2023-08-24
submit config provider

Ducica:
@mirekys Ahoj Miro, as discussed, I have implemented a context that enables you to configure our useSubmit hook before making form submission. Example button that uses it

`import React from "react";
import { Button } from "semantic-ui-react";
import { i18next } from "@translations/docs_app/i18next";
import { useFormikContext } from "formik";
import {
  useSubmitConfig,
  useFormConfig,
  submitContextType,
} from "@js/oarepo_ui";

export const SaveButton = ({ ...uiProps }) => {
  const { handleSubmit, isSubmitting } = useFormikContext();
  const { updateConfig } = useSubmitConfig();
  const { formConfig } = useFormConfig();
  return (
    <Button
      name="save"
      disabled={isSubmitting}
      loading={isSubmitting}
      color="grey"
      onClick={(e) => {
        updateConfig({
          apiUrl: formConfig.createUrl,
          context: submitContextType.create,
          onSubmitSuccess: (result) => {
            window.location.href = window.location.pathname.replace(
              "_new",
              result.id
            );
          },
        });
        handleSubmit(e);
      }}
      icon="save"
      labelPosition="left"
      content={i18next.t("Save")}
      type="button"
      {...uiProps}
    />
  );
};
`

It works well. In my opinion, the minus is that depending on what actions need to be taken before, after the submission, you could potentially have some more logic in the component that is calling the submit operation.  However, this would have to be somewhere anyway. For example, in invenio, the submit context is very slim, but it is because all of the meat is within redux actions themselves and we are not applying the same approach. Please take a look and let me know what you think.


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/util.js
```diff
@@ -56,3 +61,48 @@ export const eliminateUsedLanguages = (
...
+  const processArray = (arr) => {
+    return arr.map((item) => {
+      if (_isObject(item)) {
+        return _omit(item, "__key");
+      }
+      return item;
+    });
+  };
+
+  const processObject = (obj) => {
```
mirekys, line 75: What happens when there is a `__key` key in the top-level `inputObject`? It doesn't seem to me that it gets removed here. It removes `__key` only from child objects present in array-typed values only. Additionaly, cannot the `removeNullAndUnderscoreProperties` also take care of this stuff? :) It is just an another kind of underscore prop

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/util.js
```diff
@@ -56,3 +61,48 @@ export const eliminateUsedLanguages = (
...
+// from internal arrayField (i.e. subjects field) which is solvable, but better to wait and see how we will
+// actually hande this issue in reality before
+export const removeNullAndUnderscoreProperties = (values, formik) => {
+  const newValues = _omitBy(
+    values,
+    (value, key) =>
+      value === null ||
+      (Array.isArray(value) && value.every((item) => item === null)) ||
+      key.startsWith("_") ||
+      key === "revision_id" ||
```
mirekys, line 101: I would avoid hardcoding specific fields here. Let's name it `removeNullAndInternalFields`, and then remove every field prefixed by "_", and anything passed in a list of internalFields as function argument

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/util.js
```diff
@@ -56,3 +61,48 @@ export const eliminateUsedLanguages = (
...
+      key === "links" ||
+      key === "updated" ||
+      key === "created"
+  );
+  newValues.metadata = removeKeyFromNestedObjects(newValues.metadata);
```
mirekys, line 106: This `.metadata` breaks with vocabularies use-case (there is no metadata key in values), we shouldn't assume much here about the structure of values :)

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/contexts.js
```diff
@@ -1,8 +1,54 @@
...
+
+  return (
+    <SubmitConfigContext.Provider value={contextValue}>
+      {children}
+    </SubmitConfigContext.Provider>
+  );
+};
+
+export function useSubmitConfig() {
+  return useContext(SubmitConfigContext);
```
mirekys, line 53: Check here before returning context, that we use the hook inside context provider, otherwise, throw an error for the developer

mirekys (changes requested):

Ducica:
@mirekys Ahoj I made the requested modifications. Some notes:

Function removeNullAndInternalFields takes in internalFieldsArray and keysToRemove array.

Then it takes the formik's values (i.e. our current record) and first strips some items from it i.e. items that begin with _ or some other fields that we don't need for submitting such as created_at and so on. We are talking about top level items here.

Then removeKeyFromNestedObjects function takes this object, goes through it and deletes all the keys provided in the keysToRemove array that can be in arbitrarily nested structure underneath. And then a clean object that can be submitted is returned.

mirekys (approved):

Ducica:
@mirekys @mesemus Zdar,

I am tagging both, as we were all involved in this subject. The changes we discussed yesterday are implemented in this PR.

The main things are happening in client.js (low level API), depositAction.js (higher level with actions) - I made this class kind of flat with simple actions. I don't know if we want to have another layer of abstraction included somehow - please comment. Functions all work and if I may say so, better than before. The apiClient context provider is in the contexts.js module. I have tried including a lot of comments to explain my reasoning and I would like to hear if I did not get something well. To see the functionality, there is a branch in NR - docs stojanovic/fe-12-implement-react-depositedit-form https://github.com/Narodni-repozitar/nr-docs/tree/stojanovic/fe-12-implement-react-depositedit-form

Problems:

- I don't see where reasonably to include useMutation hook from react query. I feel like it is not needed for these specific types of actions for deposits, but we could still use it for getting data from APIs, which we will definitely be doing.
- this new approach breaks vocabularies, because they rely on this useSubmitHook, so I am not sure how we wish to handle this
-  Error handling is not very deeply implemented yet, but I checked and it propagates back to the formik's state - this needs more work (this partly needs to be done in NR-Docs and partly in OAREPO-ui). I also don't know how detailed we wish to go with 400 and 500 errors.


Additional problems (I would say BE related):
- I opened BE112 in linear, sometimes the links provided within the record work in a weird way, but when I strip the protocol, hostname and port, they start to work. The problem seems kind of random though, meaning sometimes the regular link would work and sometimes not - not sure if I am the problem or if there is some other issue.
- I am not sure how record will be actually coming from BE. Currently, metadata comes as empty object (in case of new) and this is not optimal. This is generally  a part of a bit deeper problem. For example, invenio uses some serializers deserializers, which are pretty big chunks of code that are very hardcoded and set the initial data of fields to some values and so on. In our case, because of this, we can also be getting error "changing uncontrolled to controlled input". But I think for some nested array fields, this would not be possible to avoid anyway (as it is not known ahead of time if these inputs will exist or not). Also, I am not sure if we additionally need to strip all empty fields i.e. if date is empty to not try to send empty field, but to not send anything. 
- also FE 50 issue I am getting in console - seems to be some template inheritance issue


Thanks a lot :))

mirekys (changes requested):

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/client.js
```diff
@@ -8,6 +8,9 @@
 import axios from "axios";
 import _get from "lodash/get";
 
+// create URL is fixed and gotten from the HTML, it would be good to code it straight into the API client
```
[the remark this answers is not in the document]
Ducica: @mirekys Ahoj, I did not want to duplicate getinput from DOM in client.js, so I decided to just use useFormConfig inside APiClientProvider, so we don't really need to do anything about it I would say

`  const {
    formConfig: { **createUrl }**,
  } = useFormConfig();
  // DefaultClass muze byt to OARepo.cosi..Client co ted mame
  const formik = useFormikContext();
  // initializing the low level api client with createUrl - the rest of URLs come from the record.links
  const OARepoFormActions = useMemo(() => {
    return new clientClass(new ApiClient(**createUrl)**, formik);
  }, [createUrl, formik, clientClass]);`

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/client.js
```diff
@@ -78,43 +81,108 @@ export class DepositApiClient {
...
    * @param {object} draft - Serialized draft
    */
-  async createDraft(url, payload) {
-    return this._createResponse(() => this.axiosWithConfig.post(url, payload));
-  }
+  createDraft = async (draft) => {
+    return this._createResponse(() =>
+      this.axiosWithConfig.post(
+        this.createUrl,
+        // this.createUrl.replace("https://0.0.0.0:5000", ""),
```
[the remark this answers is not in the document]
Ducica: @mirekys For sure, but we have a small problem, because I don't think we want to do this stripping in code in reality, but if I don't strip it I get that error from https://linear.app/ducesnet/issue/BE-112/issues-with-content-security-policy sometimes. (when I use the links as they come in record, without stripping protocol,host and port). But the issue is kind of random and I don't understand why it is happening.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/client.js
```diff
@@ -78,43 +81,108 @@ export class DepositApiClient {
...
+   * @param {object} draft - the draft payload
+   */
+  saveDraft = async (draft) => {
+    return this._createResponse(() =>
+      this.axiosWithConfig.put(
+        draft.links.self,
+
+        // draft.links.self.replace("https://0.0.0.0:5000", ""),
+        // "/fake/api",
+        draft
```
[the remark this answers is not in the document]
Ducica: @mirekys I can discuss with Miro for sure. I tried using this addition in our API calls in one of previous iterations and to be honest, I did not notice any difference in the response returned. I am not sure if something on the BE is not the same - so currently I don't understand what this actually does :)

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/client.js
```diff
@@ -78,43 +81,108 @@ export class DepositApiClient {
...
-  async createDraft(url, payload) {
-    return this._createResponse(() => this.axiosWithConfig.post(url, payload));
-  }
+  createDraft = async (draft) => {
+    return this._createResponse(() =>
+      this.axiosWithConfig.post(
+        this.createUrl,
+        // this.createUrl.replace("https://0.0.0.0:5000", ""),
+        // "/fake/api",
+        draft
```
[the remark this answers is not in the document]
mirekys: This includes fields marked in back-end as expandable, and adds them to top-level record metadata under `expanded` key, like here: https://github.com/inveniosoftware/invenio-rdm-records/blob/2e363f35c1df91a8ba92b28a7fb26035b857d976/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.test.js#L121 , we don't need it now, but we may come back to it in the future (with drafts in communities & requests)

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/depositActions.js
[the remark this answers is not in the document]
Ducica: @mirekys client.js contains saveOrPublish method that calls createDraft in case id is not existing, and calls saveDraft in case of existing id

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
```diff
@@ -22,45 +21,87 @@ export const useVocabularyOptions = (vocabularyType) => {
...
+export const useConfirmationModal = () => {
+  const [isModalOpen, setIsModalOopen] = React.useState(false);
+
+  const handleCloseModal = () => setIsModalOopen(false);
+  const handleOpenModal = () => setIsModalOopen(true);
+
+  return { isModalOpen, handleCloseModal, handleOpenModal };
+};
+
+export const submitContextType = {
```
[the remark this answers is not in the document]
Ducica: @mirekys probably not, I kept it for the time being together with useOnSubmit, as we are not yet 100% sure on how we will proceed

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/depositActions.js
```diff
@@ -0,0 +1,199 @@
...
+export class DepositActions {
+  constructor(initializedApiClient, formik) {
+    this.apiClient = initializedApiClient;
+    this.formik = formik;
+    this.isSubmitting = formik.isSubmitting;
+  }
+
+  async save() {
+    let response;
+    const cleanedValues = removeNullAndInternalFields(
```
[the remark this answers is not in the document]
Ducica: @mirekys we discussed this before, and we were not sure how we wish to manage serialization/deserialization, as invenio's class is very hardcoded and we did not like this approach, so we put a pin in it, but I am not sure how we wish to do it. Generally, in invenio, they deserialize record on arrival in the app from html  and placed in store, and then later in API client (our client.js).
They serialize before shipping and deserialize the response like this

` constructor(additionalApiConfig, createDraftURL, recordSerializer) {
    super(additionalApiConfig);
    this.createDraftURL = createDraftURL;
    this.recordSerializer = recordSerializer;
  }

  async _createResponse(axiosRequest) {
    try {
      const response = await axiosRequest();
      const data = **this.recordSerializer.deserialize(response.data || {});**
      const errors = this.recordSerializer.deserializeErrors(
        response.data.errors || []
      );
      return new DepositApiClientResponse(data, errors);
    } catch (error) {
      const errorData = error.response.data;
      throw new DepositApiClientResponse({}, errorData);
    }
  }`

` async saveDraft(draft, draftLinks) {
    const payload = **this.recordSerializer.serialize(draft);**
    return this._createResponse(() =>
      this.axiosWithConfig.put(draftLinks.self, payload, {
        params: { expand: 1 },
      })
    );
  }`
  
  So we could probably do something similar, only I don't know how we want this serializer class to look exactly, as they are using these Field, SchemaField and so on, with hardcoded fieldPaths that match RDM requirements.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/client.js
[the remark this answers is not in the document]
mirekys: Yep I agree on that, we should fix in BE-112 asap, so that it provides proper links, we shouldn't neet to modify them here in FE

mirekys:
Hi all,

> Problems:
> 
> * I don't see where reasonably to include useMutation hook from react query. I feel like it is not needed for these specific types of actions for deposits, but we could still use it for getting data from APIs, which we will definitely be doing.

useMutation is not needed right now. When needed, a place for it to live should be in either deposit actions client or low-level API client, I think. But that would mean that it couldn't be a class (classes cannot use any React hooks).

> * this new approach breaks vocabularies, because they rely on this useSubmitHook, so I am not sure how we wish to handle this

To fix this, vocabularies forms needs to be updated to use the same pattern as drafted here (wrap form \w ApiContextProvider, get API client from context & use client actions on form submit event)

> * Error handling is not very deeply implemented yet, but I checked and it propagates back to the formik's state - this needs more work (this partly needs to be done in NR-Docs and partly in OAREPO-ui). I also don't know how detailed we wish to go with 400 and 500 errors.
> 

Here we can provide atleast some basic error handling in low-level API client, and register a base record serializer class,
with generic implementation of `deserializeErrors` function to provide error state compatible with Formik.

> Additional problems (I would say BE related):
> 
> * I opened BE112 in linear, sometimes the links provided within the record work in a weird way, but when I strip the protocol, hostname and port, they start to work. The problem seems kind of random though, meaning sometimes the regular link would work and sometimes not - not sure if I am the problem or if there is some other issue.

For this, as a better workaround before BE-112 gets fixed, we won't do a string replace, but rather extract the URL path + query from record links and use that.

> * I am not sure how record will be actually coming from BE. Currently, metadata comes as empty object (in case of new) and this is not optimal. This is generally  a part of a bit deeper problem. For example, invenio uses some serializers deserializers, which are pretty big chunks of code that are very hardcoded and set the initial data of fields to some values and so on. In our case, because of this, we can also be getting error "changing uncontrolled to controlled input". But I think for some nested array fields, this would not be possible to avoid anyway (as it is not known ahead of time if these inputs will exist or not). Also, I am not sure if we additionally need to strip all empty fields i.e. if date is empty to not try to send empty field, but to not send anything.

This will need to get fixed in BE-129, for now we should assume here, that we will receive all fields (including any nested fields set to its default value according to field type), for array-type fields, they will receive a value of `[]`.

We will manage any empty fields upon form submission, a record serializer registered on low-level API client should be responsible for that task.

> * also FE 50 issue I am getting in console - seems to be some template inheritance issue
>

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/client.js
```diff
@@ -78,43 +81,108 @@ export class DepositApiClient {
...
    *
    * @param {object} draft - Serialized draft
    */
-  async createDraft(url, payload) {
-    return this._createResponse(() => this.axiosWithConfig.post(url, payload));
-  }
+  createDraft = async (draft) => {
+    return this._createResponse(() =>
+      this.axiosWithConfig.post(
+        this.createUrl,
```
mirekys, line 106: This is the only place createURl is needed, it would make sense to me to move it from instance to this function params

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/client.js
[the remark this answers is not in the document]
mirekys: I suggest to move it to createDraft func params, see comment bellow, which would make it a responsibity of that function caller, which is always a form that has, and should have access to that url by form config. This api client shouldnt depend on form config to be able to create an instance of itself

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/contexts.js
```diff
@@ -1,8 +1,47 @@
...
-}
+export const ApiClientContext = React.createContext();
+
+export const ApiClientProvider = ({
+  children,
+  clientClass = DepositActions,
+}) => {
+  const {
+    formConfig: { createUrl },
+  } = useFormConfig();
```
mirekys, line 25: We shouldnt be depending on access to form config here anymore, in order to provide api client to children. there might be some non-formik based apps needing to access the apis

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/contexts.js
```diff
@@ -1,8 +1,47 @@
...
+  // DefaultClass muze byt to OARepo.cosi..Client co ted mame
+  const formik = useFormikContext();
```
mirekys, line 27: same point as formConfig, here we kinda restrict any api client access to formik based children only (if this raises error elsewhere outside formik context). Would be nice if this could be moved to clientClass implementation where its actually needed and specific for current usage (form deposit actions). This also causes the following useMemo to recreate itself on each formik change

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/contexts.js
```diff
@@ -1,8 +1,47 @@
...
+  // initializing the low level api client with createUrl - the rest of URLs come from the record.links
+  const OARepoFormActions = useMemo(() => {
+    return new clientClass(new ApiClient(createUrl), formik);
```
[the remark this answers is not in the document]
mirekys: I suggest creating & passing baseClientClass param with class name to constructor & create base api client instance there

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/client.js
Ducica: @mirekys Ahoj Miro, firstly, thanks a lot for all the detailed comments you have made. I have already resolved most of the stuff, but for a few items, I have some problems. Initially, I was passing the createUrl as a param to createDraft method, but it turned out to be kind of problematic.

For example, the caller of createDraft, in case of NR-DOCS would be our depositActions class, that does not really have access to createUrl either. So this would mean I would kind of have to pass it from the button where I am calling the method from depositActions, to be passed down to the createDraft in client.js.

Maybe some compromise would be to add createUrl as a param to createDraft, but give it default value from the class constructor and in that way if someone chooses to use createDraft straight from the low level API class, they can provide the create URL themselves? Somehting like this:

`  constructor(createUrl = undefined) {
    super();
    this.createUrl = createUrl;
  }`


`  createDraft = async (draft, createUrl = this.createUrl) => {
    if (!createUrl)
      throw new Error(
        "You must either pass createUrl when initializing the OARepoDepositApiClient class or pass it to createDraft method., "
      );
    return this._createResponse(() =>
      this.axiosWithConfig.post(createUrl, draft)
    );
  };`

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/contexts.js
Ducica: @mirekys but formConfig is the only place from where I can gain access to createUrl currently in our form apps. Also all the relevant data necessary for the form app to work are coming from it. Probably I could get input from HTML directly, too, but this would be just duplicating what formConfig is already providing us. Aren't all of our APPs, going to have the same principle of getting record from the HTML and passing it to javascript?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/contexts.js
Ducica: @mirekys I fully agree with you, but I don't know how to make this happen. In current implementation, we specifically placed ApiClientProvider in DepositForm in location below BaseForm, so that I could call useFormikContext inside and pass it to the DepositActions (apiClient) class. DepositActions class, in its current implementation is 100% reliant on formik and cannot do much without it. Since I don't  see a way to gain access to formik directly from DepositActions class, I can either pass it here as it is now, or I could technically pass it to the action (individual method of depositAction class that I would be calling from buttons) in the button component (I could call useFormikContext there and then pass formik down). Do you have some suggestions please?

Ducica:
@mirekys Ahoj Miro,

I've made the requested changes and pushed. Some notes:

1. It does not work well to pass these values that I am providing as defaults to useDepositApiClient hook internalFieldsArray = [
    "errors",
    "validationErrors",
    "httpErrors",
    "successMessage",
  ],
  keysToRemove = ["__key"], directly to the serializerClass, because in case I use the hook and don't provide any values to it, these undefined or eventually empty array internalFieldsArray and keysToRemove would override the values I set as defaults in the class itself. So I left them in the hook itself.

2. I've moved FormikStateLogger to oarepo-ui, because I think it makes more sense to be there and I will remove it from vocabularies when I refactor the code to use the new approach of posting requests.

Please take a look and let me know if OK.

Thanks a lot,

Dusan

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
```diff
@@ -22,45 +25,210 @@ export const useVocabularyOptions = (vocabularyType) => {
...
+    }
+  }
+
+  async function _delete() {
+    setSubmitting(true);
+
+    try {
+      let response = await apiClient.deleteDraft(values);
+
+      // TODO: should redirect to /me page in user dashboard?? but we don't have that one yet
```
mirekys, line 203: We should consider adding redirect url as function parameter here, there shouldn't be any hardcoded `/docs/` redirect url as this is just a generic ui library. I'll update the code here accordingly

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/util.js
```diff
@@ -56,3 +60,53 @@ export const eliminateUsedLanguages = (
   );
   return remainingLanguages;
 };
+
+// function that goes through an object and all its members and removes all the keys in the
+// list of keys provided as argument from any object that lives inside the initial object,
+//  even if it is inside of an array
+
+function removeKeyFromNestedObjects(obj, keysToRemove) {
```
mirekys, line 68: Is this func used from anywhere? (e.g. record serializer)

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
Ducica: @mirekys Ahoj, agreed, it was just a temporary measure. In RMD, they hardcode url as /me/uploads, which takes the user back to their dashboard. I was thinking that some similar link, would come inside of formConfig, as it is not exactly a link that should be delivered within the record's links. However, do I understand well, that you wish for _delete function in useDepositApiClient to accept a redirection url param and to redirect to that page after deletion is done?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/util.js
Ducica: @mirekys yes it is used in the serializer, I forgote to delete it from utils, my bad.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
mirekys: Yes, exactly, and then we can pass it whatever we want, it will be specific for each instance of form & repository. For draft-enabled repositories, it will be `/user/<record-model>` (e.g. `/user/docs/`), for vocabularies, it could be just `/vocabularies/` or parent vocabulary item detail url...

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/api/client.js
```diff
@@ -77,44 +80,97 @@ export class DepositApiClient {
...
 
   /**
    * Calls the API to read a pre-existing draft.
    *
    * @param {object} draftLinks - the draft links object
    */
-  async readDraft(draftLinks) {
-    return this._createResponse(() => this.axiosWithConfig.get(draftLinks));
-  }
+  readDraft = async (draft) => {
```
mirekys, line 144: We don't have draft data to be passed, we are trying to read them here, using draftLinks

mirekys (changes requested):

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
```diff
@@ -22,45 +25,210 @@ export const useVocabularyOptions = (vocabularyType) => {
...
+  const {
+    isSubmitting,
+    values,
+    validateForm,
+    setErrors,
+    setSubmitting,
+    setValues,
+    setFieldError,
+    setFieldValue,
+  } = useFormikContext();
```
mirekys, line 77: cant we do `const formik = useFormikCOntext` and then `const { isSubmitting, values,....} = formik` here? So that we can export the whole formik object for users of the hook?

Ducica: @mirekys You are correct, I did not think of it to do it like that. Thanks!

mirekys (approved):

PR #85, opened 2023-09-18
added hoverstate and few props

Ducica:
@mirekys Zdar, here is what we discussed today. There are no big changes, a few props and hover state possibility (when you hover the close button to highlight the inputs you will remove). Please take a look.


Ducica:
@mirekys ahoj Miro, I pushed changes like we discussed. Please take a look when you can. Thanks

mirekys (changes requested):

mirekys (approved):

PR #99, opened 2023-10-23
Stojanovic/ep 34 fix search page rendering

Ducica:
@mirekys I've made the changes to layout like we discussed. Also, I pushed the result count to the right side next to sort options - not 100% sure if we like this or not. 

Main changes are in file ActiveFiltersElement.jsx. Some notes regarding the file (as I understand it might seem a bit weird at first):

1. I brought aggregations in, so that I would have access to facet group labels
2. Only way to conditionally render things in semantic UI based on screen size seems to be by using grid columns/rows with "only" prop. Otherwise, if we don't like this, I could add some selectors and write media queries in CSS to achieve this. Or maybe there is another approach I don't know about?
3. Grid.Row is flexed by default and this made groups of facets be inline with one another and I did not like this - but not sure if this is actually a problem
image

I prefer it like on Alza:

image

4. I had alignment issues with Clear filters button on mobile/tablet, so I solved it with inline style 
image

Please take a look and let me know!


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppResultOptions.jsx
```diff
@@ -15,10 +14,8 @@ export const SearchAppResultOptions = ({ sortOptions, layoutOptions }) => {
   return (
     <Grid>
       <Grid.Row verticalAlign="middle">
-        <Grid.Column floated="left" textAlign="left" width={5}>
+        <Grid.Column textAlign="right" floated="right" width={13}>
```
mirekys, line 17: Screenshot 2023-10-25 at 10 43 08

please increase padding between elements here, or make total results look like regular text (not label)

mirekys (changes requested):

mirekys:
> @mirekys I've made the changes to layout like we discussed. Also, I pushed the result count to the right side next to sort options - not 100% sure if we like this or not.
> 
> Main changes are in file ActiveFiltersElement.jsx. Some notes regarding the file (as I understand it might seem a bit weird at first):
> 
> 1. I brought aggregations in, so that I would have access to facet group labels
> 2. Only way to conditionally render things in semantic UI based on screen size seems to be by using grid columns/rows with "only" prop. Otherwise, if we don't like this, I could add some selectors and write media queries in CSS to achieve this. Or maybe there is another approach I don't know about?
> 3. Grid.Row is flexed by default and this made groups of facets be inline with one another and I did not like this - but not sure if this is actually a problem
>    image
> 
> I prefer it like on Alza:
> 
> image
> 
> 4. I had alignment issues with Clear filters button on mobile/tablet, so I solved it with inline style
>    image
> 
> Please take a look and let me know!

Thans a lot @Ducica, I also prefer Alza layout with filter groups layed out as flex column (I saw it like that in current implementation)

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppResultOptions.jsx
Ducica: @mirekys thanks pushed

Ducica:
@mirekys Zdar, I tried implementing changes as per today's discussion, please take a look and thanks!

mirekys (approved):

mirekys (approved):

PR #108, opened 2023-11-15
Stojanovic/fe 108 find a component for dates

Ducica:
@mirekys this is the edtf datepicker component.


oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/theme/globals/site.overrides
mirekys, line 1: All these styles should be moved & encapsulated into [Semantic UI less component](https://semantic-ui.com/usage/theming.html#definition-file). Apply these styles only for childern of elements having `className="ui datepicker ..."`. Global site overrides should contain only definitions, that needs to be included on every page on every possible repository site...

mirekys (changes requested):

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/EDTFDatePickerField/EDTFDatePickerField.jsx
```diff
@@ -0,0 +1,223 @@
...
+import { Form } from "semantic-ui-react";
+import { i18next } from "@translations/oarepo_ui/i18next";
+
+const edtfDateFormatOptions = [
+  { value: "yyyy", text: i18next.t("YYYY") },
+  { value: "yyyy-MM", text: i18next.t("YYYY-MM") },
+  { value: "yyyy-MM-dd", text: i18next.t("YYYY-MM-DD") },
+];
+
+const useInitialDateFormat = (fieldValue) => {
```
mirekys, line 17: Aren't there any tools to detect date format from string, or can't we do a simple regex match here?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/EDTFDatePickerField/EDTFDatePickerField.jsx
```diff
@@ -0,0 +1,223 @@
...
+    dateFormat = "yyyy-MM-dd";
+  }
+
+  const [initialDateFormat, setInitialDateFormat] = useState(dateFormat);
+  return [initialDateFormat, setInitialDateFormat];
+};
+
+const allEmptyStrings = (arr) => arr.every((element) => element === "");
+
+const pad = (value) => (value < 10 ? `0${value}` : value);
```
mirekys, line 40: This can be done entirely by lodash: `_padStart(value, 2, '0')`

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/EDTFDatePickerField/EDTFDatePickerField.jsx
```diff
@@ -0,0 +1,223 @@
...
+
+const serializeDate = (dateObj, dateFormat) => {
+  if (dateObj === null) return "";
+
+  if (dateFormat === "yyyy") return `${dateObj.getFullYear()}`;
+  if (dateFormat === "yyyy-MM")
+    return `${dateObj.getFullYear()}-${pad(dateObj.getMonth() + 1)}`;
```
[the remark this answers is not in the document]
mirekys: What happens on December? :)

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/EDTFDatePickerField/EDTFDatePickerField.jsx
```diff
@@ -0,0 +1,223 @@
...
+  if (dateFormat === "yyyy-MM-dd")
+    return `${dateObj.getFullYear()}-${pad(dateObj.getMonth() + 1)}-${pad(
+      dateObj.getDate()
+    )}`;
+};
+
+const deserializeDate = (edtfDateString) => {
+  if (edtfDateString) {
+    return new Date(edtfDateString);
```
mirekys, line 56: How does this work with EDTF ranges like 2012/2021? :)

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/EDTFDatePickerField/EDTFDatePickerField.jsx
```diff
@@ -0,0 +1,223 @@
...
+      <Form.Input
+        className="datepicker-wrapper-SUI-input"
+        error={meta.error}
+        required={required}
+        label={<FieldLabel htmlFor={fieldPath} icon={icon} label={label} />}
+      >
+        <DatePicker
+          {...field}
+          selected={
+            deserializeDate(field?.value) ? deserializeDate(field?.value) : null
```
mirekys, line 180: Do we need to run deserialization twice here?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/EDTFDatePickerField/EDTFDatePickerField.jsx
```diff
@@ -0,0 +1,223 @@
...
+  helpText,
+  required,
+  placeholderText,
+  ...datePickerProps
+}) => {
+  const { setFieldValue } = useFormikContext();
+  const [field, meta] = useField(fieldPath);
+  const [dateFormat, setDateFormat] = useInitialDateFormat(field?.value);
+
+  const onChange = (dates) => {
```
mirekys, line 165: Is dates an array here? Or is it always a string in this case?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/EDTFDatePickerField/EDTFDatePickerField.jsx
Ducica: @mirekys Ahoj, the value coming from BE, is passed through native JS Date class because this datepicker works with Date objects and not strings. No mater in which shape the string arrives, if it can be handled by Date class, it will be displayed in the input normally, but it will fail validation if you try to save the work. 

image

I have included a try catch block that console.logs an error in case the string coming from the BE has invalid format. 

Though if this does happen to be the case, anyway, I think the only way for user to actually save or do anything with this record would be to choose the proper date via the datepicker.

mirekys (approved):

mirekys (approved):

PR #115, opened 2023-12-05
current user refresh

Alzpeta:


oarepo_ui/resources/catalog.py
```diff
@@ -20,9 +25,95 @@
 class OarepoCatalog(Catalog):
     singleton_check = None
 
-    def __init__(self):
-        super().__init__()
-        self.jinja_env.undefined = jinja2.Undefined
+    def __init__(
```
mesemus, line 28: Do we have to overwrite the constructor? It seems that the only line modified is 51 and that is done to pass the current_app.

The app is used on 101: 
`if name in self.jinja_env.app.template_context_processors`

can it be changed to `if name in current_app.template_context_processors` ?

Also, we need tests for this functionality (register test blueprint with context processor and make sure it gets called)

mesemus (changes requested):
Please see the comment in catalog.py

mesemus (approved):

PR #132, opened 2024-02-19
Token api implementation

mesemus:
This pull request adds an implementation of a token retrieval page - a pure HTTP GET page where user could 
receive a token for a commandline client.

Usage:

1. User calls a nrp client to add a repository
   `nrp add alias https://127.0.0.1:5000 blah`
2. This client redirects user to a token retrieval page, that is guarded with login_required. Parameter
    `name` is passed as query parameter to the page and will be used as token name
3. If there is already a token with this name, it is deleted
4. A new token is created and access_token is displayed to the user
5. User copies the token and pastes it to the commandline client
6. Commandline client stores the token to the config file and starts using it

### DIFF
--- a/oarepo_ui/resources/resource.py
+++ b/oarepo_ui/resources/resource.py
@@ -1,4 +1,6 @@
 import copy
+import functools
+import re
 from functools import partial
 from typing import TYPE_CHECKING, Iterator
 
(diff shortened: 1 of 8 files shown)


mirekys (commented):
We decided to use the default invenio ui for this

PR #134, opened 2024-02-24
Introduced DataField, removed old code from oarepo-ui, refactored sources

mesemus:


oarepo_ui/resources/templating/data.py
```diff
@@ -0,0 +1,107 @@
...
+    @property
+    def _is_dict(self):
+        return isinstance(self.__data, dict)
+
+    @property
+    def _is_primitive(self):
+        return self._has_value and not self._is_array and not self._is_dict
+
+
+EMPTY_FIELD_DATA = FieldData({}, {})
```
mirekys, line 107: Empty `__ui` means `_ui_label`, `_ui_hint`, `_ui_help` props would throw Python errors on access

oarepo_ui/resources/templating/data.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @staticmethod
+    def translate(x):
+        if not x:
+            return x
+        return gettext(x)
+
+    @property
+    def _ui_label(self):
+        return self.translate(self.__ui.get("label"))
```
mirekys, line 26: Missing default value when UI empty

oarepo_ui/resources/templating/data.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @property
+    def _ui_hint(self):
+        return self.translate(self.__ui.get("hint"))
```
mirekys, line 30: Missing default value when UI empty

oarepo_ui/resources/templating/data.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @property
+    def _ui_help(self):
```
mirekys, line 33: Missing default value when UI empty

mirekys (changes requested):

PR #139, opened 2024-03-02
Custom fields in detail

mesemus:


oarepo_ui/resources/templating/filters.py
```diff
@@ -63,6 +63,29 @@ def field_value(value):
...
+    if isinstance(value, FieldData):
+        return value
+    return FieldData(value, ui=ui or {})
+
+
+def field_get(value, key):
+    print(f"field_get({key})")
+    if isinstance(key, FieldData):
+        key = field_value(key)
+    print("key", key)
```
mirekys, line 76: Remove this debug print

oarepo_ui/resources/templating/filters.py
```diff
@@ -63,6 +63,29 @@ def field_value(value):
...
+    if key is None:
+        return value
+
+    if not isinstance(value, FieldData):
```
mirekys, line 80: This looks like field_data()

mirekys (dismissed):

oarepo_ui/resources/templating/filters.py
Ducica: @mirekys @mesemus Hey, I've made the changes according to how I understood the issue. The normal styling of the custom fields in detail page, looks OK to me and so I was not sure what if anything to include in some of these less components. For now, their classes are simply attached to their respective parts of the Icustomfields, and so that various levels of the component can be easily overriden, which I hope is the goal :) Please take a look

oarepo_ui/templates/components/datafields/IBaseField.jinja
```diff
@@ -1,4 +1,4 @@
-{#def label, label_class="", data_class="" #}
+{#def label, label_class="ui cf-field-label", data_class="ui cf-field-data" #}
```
mesemus, line 1: would it be possible to name the component without the cf- prefix? The reason is that the base field can be used outside the custom fields as well

oarepo_ui/templates/components/datafields/ISection.jinja
```diff
@@ -0,0 +1,10 @@
+{# def title=None, title_hidden=False #}
+
+<div class="ui cf-section">
```
mesemus, line 3: would it be possible to name the component without the cf- prefix? The reason is that the section field can be used outside the custom fields as well (and is used in mbdb)

oarepo_ui/templates/components/datafields/IValueList.jinja
```diff
@@ -1,4 +1,5 @@
 {# def #}
-<dl>
+
+<dl class="ui cf-value-list">
```
mesemus, line 3: would it be possible to name the component without the cf- prefix? The reason is that the value list can be used outside the custom fields as well (and is used in mbdb)

oarepo_ui/resources/resource.py
```diff
@@ -395,11 +411,16 @@ def create(self):
...
 
         empty_record = self.empty_record(resource_requestctx)
-        form_config = self.config.form_config(
-            identity=g.identity,
-            # TODO: use api service create link when available
-            createUrl=f"/api{self.api_service.config.url_prefix}",
+
+        # TODO: use api service create link when available
+        form_config = self._get_form_config(
+            g.identity, createUrl=f"/api{self.api_service.config.url_prefix}"
```
mirekys, line 417: Still no create link on api service? :)

oarepo_ui/templates/components/EditForm.jinja
```diff
@@ -13,9 +13,9 @@
       <div class="column">
       {%- block form_main_content %}
         {%- if form_config.updateUrl %}
-        <h1 class="mt-20" style="text-transform: capitalize">{{ record.title }}</h1>
+        <h1 class="mt-20 ui header">{{ (record.title or record.metadata.title).capitalize() if record.title or record.metadata.title else '' }}</h1>
```
mirekys, line 16: else _("<Missing title>")

oarepo_ui/templates/components/datafields/IBaseField.jinja
```diff
@@ -1,4 +1,4 @@
-{#def label, label_class="", data_class="" #}
+{#def label, label_class="ui field-label", data_class="ui field-data" #}
```
mirekys, line 1: @Ducica are you sure this works? When someone uses IBaseField and passes custom `label_class` value to it?

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/modules/accordion.variables
```diff
@@ -0,0 +1 @@
+@backgroundColor: #0377cd;
```
mirekys, line 1: This should probably stay in docs (if color is related to its branding)

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/modules/field_label.variables
```diff
@@ -0,0 +1,30 @@
...
+/*
+  The type is the type of the component as defined in semantic-ui - element, collection, view, module
+  For simple component, 'element' is the correct type
+*/
+@type: "modules";
+
+/*
+  This is the name of the component - though not required, keep it the same as the filename
+*/
+@element: "cf_field_label";
```
mirekys, line 14: just field_label

Ducica: @mirekys My bad I put wrong stuff in .variables....

oarepo_ui/resources/config.py
```diff
@@ -170,18 +173,53 @@ def search_app_config(self, identity, api_config, overrides={}, **kwargs):
...
+                        "fields": [
+                            {
+                                **field,
+                                "field": prefix + field["field"],
+                            }
+                            for field in section.get("fields", [])
+                        ],
+                    }
+                )
+        return ret
```
mirekys, line 214: Possible bug, is updated `ui` also reflected by `ret` value here?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppLayout.jsx
```diff
@@ -1,26 +1,27 @@
...
 import { ClearFiltersButton } from "./ClearFiltersButton";
+import { parseSearchAppConfigs } from "@js/oarepo_ui";
 
 const ResultOptionsWithState = withState(ResultOptions);
 
 export const SearchAppLayout = ({ config, hasButtonSidebar }) => {
   const [sidebarVisible, setSidebarVisible] = React.useState(false);
-  const { appName, buildUID } = useContext(SearchConfigurationContext);
   const facetsAvailable = !_isEmpty(config.aggs);
+  const [searchAppConfig, ...otherSearchAppConfigs] = parseSearchAppConfigs();
```
mirekys, line 23: Here we shouldn't rely on this utility (here we already are inside a specific search app, which already has its specific configuration provided through its SearchConfigurationContext). You can fetch overridableIdPrefix from that context

mirekys (changes requested):

oarepo_ui/resources/resource.py
Ducica: @mirekys Hi, as far as I see, no !:)

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/modules/accordion.variables
Ducica: @mirekys Hey, the problem is that default background color is white, so the accordion is invisible, so I think better to leave it in oarepo ui, as the user could get annoyed because he is not seeing anything :) This isn't really our branding, I just liked the color that accordions have in RDM so I decided to adopt it as default color for our accordions, as we don't really have colors defined. Is that OK?

oarepo_ui/templates/components/datafields/IBaseField.jinja
Ducica: @mirekys Hello, just tested and it works. 

image

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppLayout.jsx
Ducica: @mirekys Clear, thanks. So When I am inside of components I still use app name as before if I understand well?

oarepo_ui/resources/config.py
Ducica: @mirekys Hey, the function does work, ui is a reference to an array and later the for loop is pushing to this array.

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/modules/accordion.variables
mirekys: In that case, you're right & can leave it here :)

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppLayout.jsx
mirekys: Yep it's fine to keep using appName as it's basically this: https://github.com/oarepo/oarepo-ui/blob/main/oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/util.js#L83

Ducica: @mirekys Clear thanks

oarepo_ui/resources/config.py
mirekys: Are you sure? :) There is a pending Sonar issue about this function always returning the same value no matter what happens in the for loop

Ducica: @mirekys Zdar, I can tell you for sure that it works. However, it was written by Miro. I think sonarlint is not maybe smart enough to get the code in this instance. Because ret is set to  ret = {
            "ui": ui,
        }
        
        And ui is an array, to which the for loop later appends stuff.
        
        So it will return either ui:[], or ur :[itemsfromthefoorloop]

mirekys (approved):

PR #145, opened 2024-03-14
Stojanovic/fe 124 fix remote vocabulary select field

Ducica:
Refactoring of remote field. You can now select singe item. 

I've also made some checks in case we will want to replace some of the local vocabulary selectors with remotes, due to the slowness of the form (vocabularies pre fetching), and it does work OK in context of depositform. Though, having "featured" items might be more problematic, due to the fact that you dont have any suggestions when you are starting.

### DIFF
--- a/oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/RelatedSelectField/RelatedSelectField.jsx
+++ b/oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/RelatedSelectField/RelatedSelectField.jsx
@@ -2,8 +2,8 @@ import React from "react";
 import PropTypes from "prop-types";
 import { RelatedSelectFieldInternal } from "./RelatedSelectFieldInternal";
 import { i18next } from "@translations/oarepo_ui/i18next";
-import _reverse from "lodash/reverse";
 import { getIn, useFormikContext } from "formik";
+import _isEmpty from "lodash/isEmpty";
 
 const defaultSuggestionsSerializer = (suggestions) =>
   suggestions.map((item) => ({
@@ -38,12 +38,7 @@ export const RelatedSelectField = ({
   ...uiProps
 }) => {
   const { values } = useFormikContext();
-  // console.log(multiple);
-  // console.log(
-  //   deserializeValue
-  //     ? deserializeValue(getIn(values, fieldPath, multiple ? [] : {}))
-  //     : getIn(values, fieldPath, "")
-  // );
+
   return (
     <RelatedSelectFieldInternal
       fluid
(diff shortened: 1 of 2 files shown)


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/RelatedSelectField/RelatedSelectField.jsx
```diff
@@ -38,6 +38,12 @@ export const RelatedSelectField = ({
   ...uiProps
 }) => {
   const { values } = useFormikContext();
+  const startingSuggestions = multiple
+    ? getIn(values, fieldPath, [])
+    : _isEmpty(getIn(values, fieldPath, {}))
+    ? []
```
mirekys, line 44: Why an empty array here, if in single mode and empty value

mirekys (dismissed):

Ducica:
@mirekys this one can be reviewed please

oarepo_ui/templates/oarepo_ui/footer.html
```diff
@@ -31,7 +31,7 @@
           {%- block footer_bottom %}
           <div class="eight wide column left middle aligned">
             {%- block footer_bottom_left %} {% trans
-            link="http://inveniosoftware.org" %}{{_("Powered by")}}
+            link="http://inveniosoftware.org" %}Powered by
             <a href="{{ link }}">Invenio</a>{% endtrans %} {%- endblock
```
mirekys: We would like to have this translated

oarepo_ui/templates/components/SidebarApiLinks.jinja
```diff
@@ -1,7 +1,8 @@
 {# def links #}
 <div class="ui link list">
     {% for l in links.items() %}
-        {% if "/api" in l.1 %}
+        {# to not show draft link in published record detail page#}
```
mirekys, line 4: Does this show draft links on draft record preview page however?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/ClearFiltersButton.jsx
```diff
@@ -1,20 +1,35 @@
...
-    filters.length > 0 && (
+    filters.length > initialFilters?.length && (
       <Button
         name="clear"
         color="orange"
-        onClick={() => updateQueryState({ ...currentQueryState, filters: [] })}
+        onClick={() =>
+          updateQueryState({
+            ...currentQueryState,
+            filters: currentQueryState.filters.filter((f) =>
```
mirekys, line 28: `currentQueryState.filters` == `filters` :)

oarepo_ui/templates/oarepo_ui/footer.html
Ducica: @mirekys It is already inside of this trans block, which should allow for the translation of this section. This gets extracted to the PO file, but to be honest, when I tried providing it the translation string, it did not quite work. I am not sure if there is some catch.

oarepo_ui/templates/components/SidebarApiLinks.jinja
Ducica: @mirekys We don't have a preview page where you can see the draft version of the record at this point. Though we will definitely have it in the near future. This is more a bit of a patch, in order to not show a broken link in the detail page right now, because I accidentally saw it is not working. I think that, if possible, this should be solved in the BE code that generates the links, i.e. that different sets of links would be generated for a published record and a draft and in which case such filtering would not be needed on the FE. Maybe worth discussing this with BE guys?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/ClearFiltersButton.jsx
Ducica: facepalm :)

oarepo_ui/templates/oarepo_ui/footer.html
mirekys: You are right I missed the trans block completely 🙈

mirekys (dismissed):

oarepo_ui/templates/components/SidebarApiLinks.jinja
mirekys: This should probably be tracked in: #EP-51

oarepo_ui/templates/oarepo_ui/footer.html
mirekys: Tried to fix it but also no success. But don't think this trans block is a blocking issue :)

mirekys (approved):

PR #157, opened 2024-04-04
wrapper cmp for displaying active filters

Ducica:
I find that I am often repeating same logic in terms of displaying active filters. The active filters component and clearfilters button, have mechanism to not render if certain conditions are not met, but the problem is these components are usually nested inside some grid row//column and then I get empty column or a row in the layout, which is not OK. React searchkit has a nice base for this, ShouldRender cmp, but unfortunately, that one is not exported from the library for some reason, so I pasted it into our code. I guess it could be useful in some other contexts as well in the future.

### DIFF
--- a/oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppLayout.jsx
+++ b/oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppLayout.jsx
@@ -14,6 +14,7 @@ import {
 } from "@js/invenio_search_ui/components";
 import { ResultOptions } from "@js/invenio_search_ui/components/Results";
 import { ClearFiltersButton } from "./ClearFiltersButton";
+import { ShouldActiveFiltersRender } from "@js/oarepo_ui";
 
 const ResultOptionsWithState = withState(ResultOptions);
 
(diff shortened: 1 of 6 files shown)


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppLayout.jsx
```diff
@@ -134,19 +135,23 @@ export const SearchAppLayout = ({ config, hasButtonSidebar }) => {
...
         </Grid.Row>
-        <Grid.Row>
-          <Grid.Column floated="left">
-            <ClearFiltersButton />
-          </Grid.Column>
-        </Grid.Row>
+        <ShouldActiveFiltersRender>
+          <Grid.Row>
+            <Grid.Column floated="left">
+              <ClearFiltersButton />
```
mirekys, line 151: ClearFiltersButton also renders as <ShouldActiveFiltersRender>, isn't that duplicit?

mirekys: I get your point, in that case, I would remove it from the button itself and leave it to the creator of specific SearchAppLayout component. Layout may differ, and it could also be wrapped together with ActiveFilters under single ShouldActiveFiltersRender element

mirekys (approved):

mirekys (approved):

PR #159, opened 2024-04-08
added preview function

Ducica:


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
```diff
@@ -167,6 +167,7 @@ export const useDepositApiClient = (
   const [isSaving, setIsSaving] = useState(false);
   const [isPublishing, setIsPublishing] = useState(false);
   const [isDeleting, setIsDeleting] = useState(false);
+  const [isPreviewing, setIsPreviewing] = useState(false);
```
mirekys, line 170: There becomes way too many state variables, just to disable buttons, where there's some activity on the form. I would suggest merging all into a single state, like `isProcessing` ?

Ducica: Ahoj, initially, it was all just the isSubmitting variable from formik, but the problem with that is that I would give the loading spinner to all buttons i.e. i click on save, publish, delete, preview would get a spinner, which i found to be ugly and also potentially missleading in case the request would hang for a while. If you wish, i can put it back that way.

mirekys: I think spinners on all actions, or some spinner overlay over the whole buttons section is okay. Users shouldn't be able to click e.g. on publish, when there's some other action (like delete) already ongoing

Ducica: @mirekys Ahoj, though currently not implemented well, the idea was to use isSubmitting, to disable all the buttons via disabled prop, but for specific boolean to control if the button is loading or not. So for example, for save button, loading = {isSaving}, disabled={isSubmitting}. Anyway, if we wish to use just one boolean for all buttons, then I propose I simply use the formik's isSubmitting boolean for that, and I will remove all the other state variables (isSaving, isPublishing and so on). Just a show of what it looks like in that case

image

Please let me know and thanks!

Ducica: @mirekys Ahoj, I've made the changes as discussed, please take a look. Note: related PR https://github.com/Narodni-repozitar/nr-docs/pull/102 and publishbuttoncomponent will be removed from nr-model, leaving only the one in oarepo ui, which I think makes more sense overall.

babel.ini
```diff
@@ -21,7 +18,6 @@ encoding = utf-8
 
 [jinja2: **/templates/**.jinja]
 encoding = utf-8
-extensions = jinjax.jinjax.JinjaX
```
mirekys, line 24: This was added here by @mesemus and I think we need that

babel.ini
```diff
@@ -8,9 +8,6 @@
 
 # Extraction from Python source files
 
-[extractors]
-jinja2 = jinja2.ext:babel_extract
```
mirekys, line 12: This was added here by @mesemus and I think we need that

oarepo_ui/templates/oarepo_ui/detail.html
```diff
@@ -3,29 +3,32 @@
...
-    {% endblock %}
-    <div class="two column row">
-      {# Main content #}
-      <article
-        class="sixteen wide tablet twelve wide computer column main-record-content"
-      >
-        {% block record_main_content %}
-        {% endblock %}
-      </article>
+  <PreviewHeader is_preview={is_preview} />
```
mirekys, line 6: I would lift `is_preview` up to the detail page, and render whole PreviewHeader conditionally

babel.ini
Ducica: @mirekys This happened automatically, after using make-translations, I will put it back how it was no issue, but it would be good to know what is happening and also potentially for make-translations to not make such modifications to this file if they are actually unwanted.

oarepo_ui/translations/cs/LC_MESSAGES/messages.po
```diff
@@ -399,3 +399,28 @@ msgstr ""
...
+msgstr ""
+"Váš koncept byl uložen. Pokud chcete zobrazit jeho náhled, opravte "
+"následující chyby ověření a znovu klikněte na náhled:"
+
+msgid ""
+"Once the record is published you will no longer be able to change record's "
+"files! However, you will still be able to update the record's metadata "
+"later."
+msgstr ""
+"Jakmile bude záznam zveřejněn, již jej nebudete moci změnitsoubory! Stále "
```
mirekys, line 425: změnit soubory

oarepo_ui/translations/cs/LC_MESSAGES/messages.po
```diff
@@ -399,3 +399,28 @@ msgstr ""
...
+"však budete moci aktualizovat metadata záznamupozději."
```
mirekys, line 426: záznamu později

oarepo_ui/translations/cs/LC_MESSAGES/messages.po
Ducica: @mirekys Sorry...

mirekys (approved):

mirekys (approved):

PR #183, opened 2024-05-22
feat: update external authority modal ui/ux

mirekys:


Ducica:
The close button overlaps the search bar

image

Ducica:
Also, there is a weird issue when adding item from external authority, sequence to reproduce

choose something from external modal, then click on the affiliations input in the creatibutors modal

image

then click away, for some reason the options prop gets emptied in the SelectField, I was not able to determine why, but it only happens when item is added from external authority modal
image

PR #187, opened 2024-06-05
Stojanovic/fe 167 change form graphic per comments from ntk

Ducica:
It was not possible to use Josh's visually hidden approach, because in our case, we want the layout to be impacted by the hidden element and in his solution, the idea is that the hidden element will not impact the layout, so unfortunately it does not work well for such a use case.


mirekys:
> It was not possible to use Josh's visually hidden approach, because in our case, we want the layout to be impacted by the hidden element and in his solution, the idea is that the hidden element will not impact the layout, so unfortunately it does not work well for such a use case.

Thats okay if it's not 100% the joshes approach, as the idea was exactly to take up space, just to not be visible

mirekys (approved):

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/ArrayFieldItem/ArrayFieldItem.jsx
```diff
@@ -41,10 +50,11 @@ export const ArrayFieldItem = ({
           />
         ) : (
           <Button
-            style={{ marginTop: "1.75rem" }}
             aria-label={i18next.t("Remove field")}
             className="close-btn"
+            type="button"
             icon
+            id={indexPath}
```
mirekys, line 57: Element id should be unique across DOM, index number isn't

Screenshot 2024-06-07 at 12 30 52

mirekys (changes requested):

mirekys (approved):

mirekys (changes requested):
Screenshot 2024-06-07 at 16 36 38

`id` & `for` attrs seems to now be completely missing from DOM, & there are some unsupported attrs on invenio-group-field div

Ducica:
@mirekys Sorry about that, now it should be OK. Just for clarity, we already merged this branch to nr-docs, so it will be necessary to merge https://github.com/oarepo/oarepo-ui/pull/187 + https://github.com/Narodni-repozitar/nr-model/pull/37+ https://github.com/oarepo/oarepo-vocabularies/pull/169

Ducica:
@mirekys Sorry, I fixed it like we discussed. Unfortunately, I had to make a small change to model and vocabularies as well, 
where, for some reason I was sending this formik array to ArrayFieldItem and if you don't destructure it in the arrayFieldItem props, then it just ends up on the dom node. https://github.com/Narodni-repozitar/nr-model/pull/38/files + https://github.com/oarepo/oarepo-vocabularies/pull/170  Thanks!

mirekys (approved):

PR #190, opened 2024-06-10
Stojanovic/fe 178 validation fe

Ducica:


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
```diff
@@ -235,7 +235,7 @@ export const useDepositApiClient = ({
     }
   }
 
-  async function publish() {
+  async function publish({ validateForm = false } = {}) {
```
mirekys, line 238: Call it just validate here => no need to rename main hook prop then :)

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/elements/label.overrides
```diff
@@ -0,0 +1,17 @@
+.ui.label {
+  &.active-filter-label {
+    font-weight: normal;
+    color: black;
+  }
+  padding: 0.5rem 1rem;
```
mirekys, line 6: This has potential to break sites ui, do  we really need to do this here?

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/elements/label.overrides
```diff
@@ -0,0 +1,17 @@
...
+}
+
+.ui.labels {
+  &.active-filters-group > .label {
+    color: black;
+    margin: 0;
+    margin-right: 0.5rem;
+    padding: 0.5rem 1rem;
+  }
+  line-height: 3rem;
```
mirekys, line 16: This has potential to break ui globally

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/FoldableBucketAggregationElement.jsx
```diff
@@ -0,0 +1,44 @@
...
+          tabIndex={0}
+          active={isActive}
+          onClick={handleClick}
+        >
+          <div className="flex justify-space-between align-items-center">
+            <Header className="mb-0" as="h3">
+              {title}
+            </Header>
+            <div className="align-self-end">
+              <Icon name="angle right" />
```
mirekys, line 27: Does it rotate when activated ?

mirekys (changes requested):

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/elements/label.overrides
Ducica: @mirekys My bad, this was supposed to be inside of the active-filter-label selector. Do you generally disagree with keeping the styles here or just with its placement directly below ui.label? My logic was that, as component is in oarepo ui, it would be ideal if it did not require someone to place styles in the app itself, in order for it to look reasonable. Do you have something specific in mind please or is it enough to place it inside the active-filter-label selector?

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/elements/label.overrides
Ducica: @mirekys same as above

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/FoldableBucketAggregationElement.jsx
Ducica: @mirekys Yes, we already have css for it here https://github.com/oarepo/oarepo-ui/blob/eed9fa5abe831871ab461543cd29fb05fa3ea230/oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/modules/accordion.overrides#L4 it uses the same principle that we already have in the accordions that are part of the deposit form. I would strongly recommend actually testing out this PR by yourself please, because it would be good to be sure that things are acting exactly like we discussed and that I did not misunderstand anything.

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/elements/label.overrides
mirekys: I'd like to have it under className="ui active filter label" classes, the point here was just to avoid changing appearance of completely unrelated labels here (all "ui label" elements for everyone using oarepo-ui), if not really necessary

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/FoldableBucketAggregationElement.jsx
mirekys: Thanks, yes for sure, for the first round I was just quickly going through code on the subway :D

mirekys (changes requested):
Screenshot 2024-06-12 at 13 28 42
 Inconsistent line height after flex row wraps

mirekys (changes requested):
In mobile view, clear all filters button purpose is not very clear judging from just the icon & should be aligned left
, same with dashboard
Screenshot 2024-06-12 at 13 31 12

Ducica:
> In mobile view, clear all filters button purpose is not very clear judging from just the icon & should be aligned left , same with dashboard Screenshot 2024-06-12 at 13 31 12

@mirekys Thanks, do you mean it should go towards the menu button that opens the facets sidebar? Regarding the clarity, in the last iteration, we agreed it would be just an icon on tablet/mobile, so that it would be able to fit in the existing row with other items. Problem is if it contains text as well, it does not fit as you go smaller with screen size. Do you mean to add a popup on the icon, or do you mean that we need to display whole text please?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/ShouldActiveFiltersRender.jsx
```diff
@@ -14,8 +14,10 @@ const ShouldActiveFiltersRenderComponent = ({
     initialQueryState: { filters: initialFilters },
   } = searchAppContext;
   return (
-    <ShouldRender condition={filters.length > initialFilters?.length}>
-      {children}
+    <ShouldRender condition={filters?.length > initialFilters?.length}>
+      {React.Children.map(children, (child) =>
+        React.cloneElement(child, { filters, initialFilters })
```
mirekys, line 19: This smells like a hack :) Can't children components (the ones that actually needs it?) also directly access `currentQueryState` & `initialQueryState` through context provider ?

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/elements/button.overrides
```diff
@@ -3,4 +3,19 @@
     background: none;
     padding: 5px;
   }
-}
\ No newline at end of file
+}
+
+.ui.button {
+  &.clear-filters-button {
+    color: black;
```
mirekys, line 10: I wouldn't enforce black here, every site has different button color schemes

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/elements/label.overrides
```diff
@@ -0,0 +1,19 @@
...
+    cursor: pointer;
+  }
+  &.active-filters-count-label {
+    left: 110%;
+  }
+}
+
+.ui.labels {
+  &.active-filters-group > .label {
+    color: black;
```
mirekys, line 15: I wouldn't enforce black here, every site has different label colors

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/elements/label.overrides
```diff
@@ -0,0 +1,19 @@
+.ui.label {
+  &.active-filter-label {
+    font-weight: normal;
+    color: black;
```
mirekys, line 4: I wouldn't enforce black here, every site has different label colors

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/elements/button.overrides
Ducica: @mirekys Hey, the issue is, that, when you let semantic handle the colors, you end up with this weak contrast level
image
And I honestly, don't see another way to do it, but to just change the text color to the one that gives a good contrast value. This was done in all three places for this reason. Should I just leave the default value instead please? For me it is a bit difficult to read the text.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/ShouldActiveFiltersRender.jsx
Ducica: @mirekys I fully expected a comment here:) The problem is, initialQueryState can indeed be accessed from the context, but currentQueryState only via withState HOC. So now I have this annoying situation, where my ShouldActiveFilterRender already is wrapped with withState, and now, as its child in this case also needs it, I need to wrap it in withState again to access currentQueryState and this seemed like an antipattern i.e. they don't, make the searchkit state accessible in a very convenient way. If this is a big issue, I think I could replace this children and cloning with a render prop, where it will be something like this

`   <ShouldRender condition={filters?.length > initialFilters?.length}>
      {/* {React.Children.map(children, (child) =>
        React.cloneElement(child, { filters, initialFilters })
      )} */}
      {children({ filters, initialFilters })}
    </ShouldRender>`
    
    And then I need to replace code everywhere where I am using this component to
    
    `   <ShouldActiveFiltersRender>
              {() => (
                <ClearFiltersButton
                  className={"clear-filters-button mobile tablet only"}
                />
              )}
            </ShouldActiveFiltersRender>`
            
            To be honest, I find the current implementation nicer to use, but please let me know how you would prefer for it to be done. Thanks!

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/default/elements/button.overrides
mirekys: I now get the point here, you can leave it as is :)

mirekys (approved):

mirekys (approved):

PR #194, opened 2024-06-17
export for tiny-mce

AlbinaMakisheva:


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/util.js
```diff
@@ -61,6 +61,9 @@ const allowed_tags = [
   "ul",
 ];
 
+export const validTags =
```
mirekys, line 64: In this file, we already have a list of both `allowed_tags` and `allowedAttributes` of each tag. Can we make this a function `valid_elements` that constructs tinymce-formatted config from that info? This is to keep the number of places we need to update when the list of tags changes as low as possible

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/util.js
```diff
@@ -61,6 +61,21 @@ const allowed_tags = [
   "ul",
 ];
 
+export const validTags = (
+  tags= allowed_tags,
+  specialAttributes = {
```
mirekys, line 66: Complete duplicate of 83-87

mirekys (changes requested):

mirekys (approved):

mirekys (approved):

PR #197, opened 2024-06-27
getting allowed tags and attrs from be

Ducica:
Proposal how to propagate html tags/attrs to front end. It required also changing the function in js, because you cannot take from context inside of a javascript function. 

I also moved "calculation" of tags for the editor to the BE, because I thought it made sense for it to be calculated only one time during lifecycle of the form and all the data are already available there.

@AlbinaMakisheva Please take a look and tell me if you think this is OK

@mirekys I am not sure about crowding the resource in this way. It would be cleaner to do this via a component, but the issue is, we don't have any components used in the config on the level of oarepo ui, so if I would add a component for adding this to form config and used it in config in oarepo ui, then I would need to modify in all apps, that components first call any potential components from the parent config, before applying the app specific components to config i.e. something like this in nr-docs

class DocumentsResourceConfig(RecordsUIResourceConfig):
    template_folder = "templates"
    url_prefix = "/docs/"
    blueprint_name = "documents"
    ui_serializer_class = "documents.resources.records.ui.DocumentsUIJSONSerializer"
    api_service = "documents"

    components = [
        *RecordsUIResourceConfig.components, (use upstream cmps first)
        BabelComponent,
        DepositVocabularyOptionsComponent,
        PermissionsComponent,
        FilesComponent,
    ]

### DIFF
--- a/oarepo_ui/resources/components.py
+++ b/oarepo_ui/resources/components.py
@@ -275,22 +275,6 @@ class FilesComponent(UIResourceComponent):
 
 
 class AllowedHtmlTagsComponent(UIResourceComponent):
-    def calculate_valid_tags_for_editor(self, tags, attr):
-        special_attributes = {key: "|".join(value) for key, value in attr.items()}
-
-        result = []
-
-        if "*" in special_attributes:
-            result.append(f"@[{special_attributes['*']}]")
-
-        for tag in tags:
-            if tag in special_attributes:
-                result.append(f"{tag}[{special_attributes[tag]}]")
-            else:
-                result.append(tag)
-
-        return ",".join(result)
-
     def form_config(self, *, form_config, **kwargs):
         form_config["allowedHtmlTags"] = current_app.config.get(
             "ALLOWED_HTML_TAGS", ALLOWED_HTML_TAGS
@@ -298,8 +282,4 @@ class AllowedHtmlTagsComponent(UIResourceComponent):
 
         form_config["allowedHtmlAttrs"] = current_app.config.get(
             "ALLOWED_HTML_ATTRS", ALLOWED_HTML_ATTRS
-        )
-
-        form_config["validEditorTags"] = self.calculate_valid_tags_for_editor(
-            form_config["allowedHtmlTags"], form_config["allowedHtmlAttrs"]
         )
--- a/oarepo_ui/resources/components.py
+++ b/oarepo_ui/resources/components.py
@@ -5,6 +5,8 @@ from flask_principal import Identity
 from invenio_i18n.ext import current_i18n
 from invenio_records_resources.services.records.results import RecordItem
 from oarepo_runtime.datastreams.utils import get_file_service_for_record_service
+from invenio_config.default import ALLOWED_HTML_TAGS, ALLOWED_HTML_ATTRS
+
 
 from ..proxies import current_oarepo_ui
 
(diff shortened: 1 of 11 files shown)


mirekys:
> @mirekys I am not sure about crowding the resource in this way. It would be cleaner to do this via a component, but the issue is, we don't have any components used in the config on the level of oarepo ui, so if I would add a component for adding this to form config and used it in config in oarepo ui, then I would need to modify in all apps, that components first call any potential components from the parent config, before applying the app specific components to config i.e. something like this in nr-docs
> 
> class DocumentsResourceConfig(RecordsUIResourceConfig): template_folder = "templates" url_prefix = "/docs/" blueprint_name = "documents" ui_serializer_class = "documents.resources.records.ui.DocumentsUIJSONSerializer" api_service = "documents"
> 
> ```
> components = [
>     *RecordsUIResourceConfig.components, (use upstream cmps first)
>     BabelComponent,
>     DepositVocabularyOptionsComponent,
>     PermissionsComponent,
>     FilesComponent,
> ]
> ```

Hi Dusan, thanks for the proposal, I would vote for the component idea, as it is exactly the way of preventing
the generic resource to become bloated. There is a similar case for configuring allowed file types on a record, which is done also as a resource component

oarepo_ui/resources/components.py
```diff
@@ -270,3 +272,34 @@ def before_ui_edit(self, *, api_record, extra_context, identity, **kwargs):
 
     def before_ui_detail(self, **kwargs):
         self.before_ui_edit(**kwargs)
+
+
+class AllowedHtmlTagsComponent(UIResourceComponent):
+    def calculate_valid_tags_for_editor(self, tags, attr):
```
mirekys, line 278: I would leave this utility function purely in JS (format it outputs is relevant only to a concrete JS widget and can be easily derived from other 2 config properties, and changed, whenever editor widget changes)

oarepo_ui/resources/resource.py
```diff
@@ -47,6 +47,24 @@
     UIResourceConfig,
 )
 
+
+def calculate_valid_tags_for_editor(tags=None, attr=None):
```
mirekys, line 51: Duplicity for what is in components?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
```diff
@@ -448,3 +450,24 @@ export const useValidateOnBlur = () => {
...
+  });
+  return cleanInput;
+};
+
+export const useSanitizeInput = () => {
+  const {
+    formConfig: { allowedHtmlAttrs, allowedHtmlTags, validEditorTags },
+  } = useFormConfig();
+  return {
+    sanitizeInput: sanitizeInput(allowedHtmlTags, allowedHtmlAttrs),
```
mirekys, line 468: I would useCallback that

mirekys (changes requested):

mirekys (approved):

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
```diff
@@ -448,3 +450,29 @@ export const useValidateOnBlur = () => {
...
+      });
+      return cleanInput;
+    },
+    [allowedHtmlTags, allowedHtmlAttrs]
+  );
+
+  return {
+    sanitizeInput,
+    allowedHtmlAttrs,
+    allowedHtmlTags,
```
mirekys, line 474: why not also export a list of valid editor tags here by calling the newly created getValidTagsForEditor utility? :)

mirekys: this is to avoid constructing the list of tags in each & every component using this hook & editor

mirekys (approved):

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
```diff
@@ -448,3 +451,33 @@ export const useValidateOnBlur = () => {
...
+      const decodedString = decode(htmlString);
+      const cleanInput = sanitizeHtml(decodedString, {
+        allowedTags: allowedHtmlTags,
+        allowedAttributes: allowedHtmlAttrs,
+      });
+      return cleanInput;
+    },
+    [allowedHtmlTags, allowedHtmlAttrs]
+  );
+  const vailidEditorTags = useMemo(
```
mirekys, line 471: Looks like a typo 😅

mirekys (changes requested):

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/hooks.js
mirekys: Please fix this bit & bump the version in setup.cfg :)

mirekys (approved):

PR #210, opened 2024-08-07
Stojanovic/fe 183 validation errors translation

Ducica:
Translation of field names in FormFeedback component based labels coming from the model (with exception of files.enabled error which i solved in hardcoded way -> not sure if something can be done with this). 
The form was rerendering a lot during submitting the form (the calls to set formik state during submitting, are not in fact batched but each causes a rerender). Restructured the save hook in order to cause as little rerenders as possible. 

related PRs:

https://github.com/Narodni-repozitar/nr-model/pull/55

https://github.com/oarepo/oarepo-vocabularies/pull/178


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/FormFeedback/FormFeedback.jsx
```diff
@@ -43,25 +44,42 @@ CustomMessage.propTypes = {
...
+            const label = getFieldData({
+              fieldPath: error.field,
+              fieldRepresentation: "text",
+            })?.label;
+            return (
+              <Message.Item
+                onClick={() => scrollToElement(error.field)}
+                key={`${error.field}-${index}`}
+              >{`${
+                label
```
mirekys, line 72: refactor to `label || ...`

PR #223, opened 2024-08-21
added components and restructured styles

Ducica:


oarepo_ui/templates/components/SeparatedProperty.jinja
mirekys, line 1: Consider moving into templates/components/datafields, maybe implementing as prop of IValueList?

oarepo_ui/templates/components/Separator.jinja
mirekys, line 1: This component is the same as just directly using separator class on something, with only added value that it optionally displays nothing (given a prop that needs to be usually deeply prop-drilled). I would vote for removing this.

PR #230, opened 2024-09-09
fixed issue with rich editor

Ducica:
The PR to fix the problem with abstract field. I have followed the solution used in react-invenio-forms https://github.com/inveniosoftware/react-invenio-forms/blob/42b93636be95177f1d7845e4f55c8bb63a1d9b92/src/lib/forms/RichInputField.js#L16

image

Which is a hacky solution and causes console errors, as this prop is not supposed to be a function. It even causes an uncaught type error, as the library inside tries to strip white text from the value. This amazingly does not crash the app (i don't understand why). This is screenshot from zenodoo (live):

image

On the other hand, the problem is deeper as it is not recommended to make the editor actual controlled input, which is the paradigm that we use and is used most often in react (https://www.tiny.cloud/docs/tinymce/latest/react-ref/#using-the-tinymce-react-component-as-a-controlled-component). 

To summarize, it works, but it is not pretty and I don't like it. Alternative would be to make it actual controlled input with onChange handler, which could be inefficient in case someone writes a lot of text into the input.

Please take a look. If it is ok, I can use this component in restauration repo as well.

### DIFF
--- a/oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/I18nRichInputField/I18nRichInputField.jsx
+++ b/oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/I18nRichInputField/I18nRichInputField.jsx
@@ -3,8 +3,9 @@ import {
   LanguageSelectField,
   useSanitizeInput,
   useFieldData,
+  OarepoRichEditor,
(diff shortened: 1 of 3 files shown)


mirekys (approved):
Thanks @Ducica, let's keep it on line with Invenio's little hacky solution for now

PR #235, opened 2024-09-27
Reworking requests

Ducica:
@mirekys Here is the PR finally for requests. It consists of this PR +

https://github.com/oarepo/oarepo-requests/pull/63

https://github.com/Narodni-repozitar/nr-docs/pull/168


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/util.js
```diff
@@ -273,3 +274,24 @@ export const goBack = (fallBackURL = "/") => {
...
+  headers: {
+    Accept: "application/vnd.inveniordm.v1+json",
+    "Content-Type": "application/json",
+  },
+};
+
+export const http = axios.create(baseAxiosConfiguration);
+
+export const encodeUnicodeBase64 = (str) => {
+  return btoa(encodeURIComponent(str));
```
mirekys, line 292: SHouldn't be as `window.btoa` ?

Ducica: @mirekys Hey, I think this function is just part of the global namespace, so window. before it is optional. It works like this (without the window.)
image

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/BaseFormLayout/BaseFormLayout.jsx
```diff
@@ -32,6 +38,35 @@ export const BaseFormLayout = ({ formikProps }) => {
...
+
+  const urlHash = window.location.hash.substring(1);
+  let errorData;
+  if (urlHash) {
+    const decodedData = decodeUnicodeBase64(urlHash);
+    errorData = JSON.parse(decodedData);
+    window.history.replaceState(
+      null,
+      null,
+      window.location.pathname + window.location.search
```
mirekys, line 66: This replacing of window history feels hacky, why do we need to do this in case of any random error?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/util.js
```diff
@@ -200,3 +201,29 @@ export const getValidTagsForEditor = (tags, attr) => {
...
+  errors,
+  message = i18next.t(
+    "Draft saved with validation errors. Fields listed below that failed validation were not saved to the server"
+  )
+) => {
+  if (errors?.length > 0) {
+    let errorsObj = {};
+    const errorPaths = [];
+    for (const error of errors) {
+      errorsObj = setIn(errorsObj, error.field, error.messages.join(" "));
```
mirekys, line 215: Since this comes from user-input url #aaaaa  string, are messages sanitized before displaying in form (no-html/script should be allowed there)?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/BaseFormLayout/BaseFormLayout.jsx
mirekys: If thats the case, I would go for navigating to #, using `navigate('#');`.
This way it won't change URL and query params and it will allow you to reset the hash.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/util.js
Ducica: This function just serves to package the errors received from the server into a shape that is useful for formik's errors and for the FormFeedback cmp. The errors are received from the server and they just contain the field path for the field that resulted in error and the error message (which is array of strings) (provided by BE). No user input exists there. If I misunderstood the question, we can do a call

mirekys (approved):

mirekys:
Wrong translation, better is "Předmět žádosti"
Screenshot 2024-10-23 at 16 38 10

mirekys (dismissed):

mirekys (approved):

PR #239, opened 2024-10-14
possible to not show links in community selector:

Ducica:
After playing with the component for a while, I agree that it can be annoying for the user if they wish to select the community to be taken elsewhere. It can be even worse if community name is long.

In the PR I have removed links all together from the selector, because I simply don't think it makes sense to show for example link to the community website, if you are unable to click on it. I think it would be confusing and misleading to have it there. Please take a look and let me know what you think. Thanks!


mirekys (approved):

mirekys:
LGTM, thanks! :)

PR #265, opened 2025-01-21
feat(jinja): webpack_optional global helper

mirekys:
This allows for optional inclusion of webpack manifest:

```jinja
{{ webpack_optional('myentry') }}
```

If requested manifest not found in manifest.json, no hard error is raised.
Instead, error gets just output to resulting markup as a plain comment:

Screenshot 2025-01-21 at 17 14 53


Ducica:
@mirekys It looks good to me, after we should just make sure to remove all the current overridable registry implementation including the entry point.

Ducica (approved):

PR #272, opened 2025-02-06
feat: add record versions component and integrate into detail page

corovcam:


mirekys (commented):
Isn't there now a possibility to include https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/index.js entrypoint, and provide just the versions div in our detail template? We should already have js assets from invenio-app-rdm available to us, there shouldn't be a need to copy them over to our codebase anymore for this case

mirekys (dismissed):

mesemus:
> Isn't there now a possibility to include https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/landing_page/index.js entrypoint, and provide just the versions div in our detail template? We should already have js assets from invenio-app-rdm available to us, there shouldn't be a need to copy them over to our codebase anymore for this case

Not yet as the ui serialization needed differs

mesemus (approved):

PR #274, opened 2025-02-16
added toggler for all versions

Ducica:
It was necessary to change a bit logic in the active filters component, as some unwanted filters for which we don't have label would start appearing i.e. this allversions filter. Modified the logic of these components (active filters, clearbutton, the floating label), to look at what comes in aggs (from search config in html) and then ignore the filters that are not in aggs. Meaning, the facets on the left hand side, are being rendered from the aggs that come in search config and I DO want those to be displayed and cleared, but others not. I tested it accross the search apps and logic seems to work OK.


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppLayout.jsx
```diff
@@ -23,14 +23,17 @@ export const ActiveFiltersCountFloatingLabelComponent = ({
   className,
 }) => {
   const { filters } = currentQueryState;
-  const searchAppContext = useContext(SearchConfigurationContext);
-  const {
-    initialQueryState: { filters: initialFilters },
-  } = searchAppContext;
+  const { aggs } = useContext(SearchConfigurationContext);
```
mirekys, line 26: Would be nice it this could be hook `useActiveFilters()` as this pattern already repeats in few places

mirekys (commented):
I would extract&reuse code for getting active filters as a dedicated hook

mirekys (approved):

PR #280, opened 2025-02-24
Implement statistics section into detail sidebar

corovcam:


oarepo_ui/resources/templating/filters.py
```diff
@@ -96,3 +98,31 @@ def ijoin_filter(values, separator):
         else:
             ret.append(str(val))
     return separator.join(ret)
+
+
+def localize_number(value):
```
mirekys, line 103: Don't copy paste Invenio app RDM implementation

oarepo_ui/config.py
```diff
@@ -13,6 +13,9 @@
     "type": "oarepo_ui.resources.templating.filters:type_filter",
     "keys": "oarepo_ui.resources.templating.filters:keys_filter",
     "ijoin": "oarepo_ui.resources.templating.filters:ijoin_filter",
+    "compact_number": "oarepo_ui.resources.templating.filters:compact_number",
```
mirekys, line 16: Import RDM filters directly: https://github.com/inveniosoftware/invenio-app-rdm/blob/v12.0.13/invenio_app_rdm/records_ui/views/__init__.py#L29-L45

as `invenio_app_rdm.records_ui.filters:...

corovcam: Thanks, fixed in subsequent commit.

mirekys (approved):

PR #289, opened 2025-03-11
Stojanovic/fe 345 need reusable copy action for react part we have for jinja

Ducica:
Added copy button implementation that works both in React + jinja. 

Moved identifiers field from nr-model, as it turns out that you could need it outside of the context of "nr". For example, some specialized vocabulary items have identifiers as well that also have shape scheme + identifier. 

I will remove from nr-model + refactor other places where it is used when making refactoring to nr-docs form (for implementing RDM funders/creatibutors) in order to not make things more complicated now with additional PRs. 

PR in nr model https://github.com/Narodni-repozitar/nr-model/pull/100 (just using the new  button for copying).


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/components/ClipboardCopyButton.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+  useEffect(() => {
+    const copyBtn = copyBtnRef.current;
+    if (copyBtn) {
+      initCopyButtons(copyBtnRef.current);
+    }
+    return () => {
+      deinitializeCopyButtons(copyBtn);
+    };
+  }, [copyText]);
```
mirekys, line 17: copyText probably unnecessary dependency?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/components/ClipboardCopyButton.jsx
```diff
@@ -0,0 +1,36 @@
...
+import { i18next } from "@translations/oarepo_ui/i18next";
+import PropTypes from "prop-types";
+
+export const ClipboardCopyButton = ({ copyText, ...rest }) => {
+  const copyBtnRef = useRef(null);
+
+  useEffect(() => {
+    const copyBtn = copyBtnRef.current;
+    if (copyBtn) {
+      initCopyButtons(copyBtnRef.current);
```
mirekys, line 12: replace with copyBtn

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/components/ClipboardCopyButton.jsx
Ducica: You're right thanks!

mesemus (approved):

mesemus (approved):

PR #292, opened 2025-03-20
Display tombstone info

xsenyaaax:
Display tombstone info in table format (removal reason, note, citation text) of deleted record. Not sure about jinja template, specifically if table is written correctly


Ducica:
@xsenyaaax Hi, are there other PRs that are related to this feature please i.e. do I need anything else to test it out?  Because when I try I get this 
image

Also, how would one actually pass include_deleted to _get_record, as it is not being externally passed to any of the view functions.

xsenyaaax:
@Ducica Yes, right now there supposed to be a PR in oarepo-requests, thats adds tombstone information when published record is deleted. But after merging  with main branch, some tests failed. I need to fix them first in oarepo-requests. Once fixed I will tell you

I dont know specifically about include_deleted, Mirek Simek told me to add this flag, because invenio_rdm_records has a include_deleted when reading a record

Ducica:
> @Ducica Yes, right now there supposed to be a PR in oarepo-requests, thats adds tombstone information when published record is deleted. But after merging with main branch, some tests failed. I need to fix them first in oarepo-requests. Once fixed I will tell you
> 
> I dont know specifically about include_deleted, Mirek Simek told me to add this flag, because invenio_rdm_records has a include_deleted when reading a record

Thanks, about include_deleted, I just meant, it will always be false on all view functions (detail, form etc.), because there is no possibility to set it to true, so I am not sure if by default it should always be true, or if it is supposed to be sent via url params or smth, maybe discuss with Miro. It is always possible, that I am completely wrong :0

oarepo_ui/resources/resource.py
```diff
@@ -697,13 +706,24 @@ def expand_search_links(self, identity, pagination, args):
         return tpl.expand(identity, pagination)
 
     def tombstone(self, error, *args, **kwargs):
+        record_tombstone = error.record.get('tombstone')
```
mesemus, line 709: imho this will not work: if the record has been fetched, we are not in this error handler.

Instead, I'd call 

```
record = self._get_record(resource_requestctx, include_deleted=True)
record_tombstone=...
```

mesemus:
> Thanks, about include_deleted, I just meant, it will always be false on all view functions (detail, form etc.), because there is no possibility to set it to true, so I am not sure if by default it should always be true, or if it is supposed to be sent via url params or smth, maybe discuss with Miro. It is always possible, that I am completely wrong :0

include_deleted should be set to False in all views except tombstone view - see comment in the code

oarepo_ui/resources/resource.py
xsenyaaax: It was working for me. I went just to /docs/{deteled_record_id}, and in tombstone function there were empty args, kwargs, only defined error with deleted record. By searching in oarepo-ui files there is no other usages of this tombstone function.  

With resource_requestctx I get "RuntimeError: Working outside of resource request context."

mesemus (approved):

mesemus (approved):

PR #294, opened 2025-03-26
Add month and year dropdown components to EDTF date picker

corovcam:


Ducica:
@corovcam Thanks it looks looks/works good. What would be nice, if it is possible, to allow someone to type in the year into the year dropdown, so it is easier to find the year you want. Because when the dropdown opens, it is a bit narrow and so it would be fairly useful if I could type 202 to narrow down the selection.

Also, please correct the sonar lint warnings and do a version bump in setup.cfg

corovcam:
> @corovcam Thanks it looks looks/works good. What would be nice, if it is possible, to allow someone to type in the year into the year dropdown, so it is easier to find the year you want. Because when the dropdown opens, it is a bit narrow and so it would be fairly useful if I could type 202 to narrow down the selection.
> 
> Also, please correct the sonar lint warnings and do a version bump in setup.cfg

Thanks. 
About the year autocomplete: I tried, but it is not that straightforward to implement, as we need another state to keep track of a partial year and then narrow DatePicker maxDate with another prop, and then somehow update the formik state, but only if the year entered is a valid year and only allow user to type numbers, not letters. 
Also another thing is that we explicitly and on purpose do not allow users to type anything into the input, we just allow them to click on the DatePicker or (now) narrow the selection with Dropdowns. If we would introduce new typing mechanism into Year-only selection, it might be confusing.

Ducica:
> > @corovcam Thanks it looks looks/works good. What would be nice, if it is possible, to allow someone to type in the year into the year dropdown, so it is easier to find the year you want. Because when the dropdown opens, it is a bit narrow and so it would be fairly useful if I could type 202 to narrow down the selection.
> > Also, please correct the sonar lint warnings and do a version bump in setup.cfg
> 
> Thanks. About the year autocomplete: I tried, but it is not that straightforward to implement, as we need another state to keep track of a partial year and then narrow DatePicker maxDate with another prop, and then somehow update the formik state, but only if the year entered is a valid year and only allow user to type numbers, not letters. Also another thing is that we explicitly and on purpose do not allow users to type anything into the input, we just allow them to click on the DatePicker or (now) narrow the selection with Dropdowns. If we would introduce new typing mechanism into Year-only selection, it might be confusing.

Thanks, I thought it would not be very simple to do. Okay it works good in my opinion and we will see if the users say something about it.

Ducica (approved):

PR #297, opened 2025-04-02
Remove error colouring from whole Subjects ArrayField

corovcam:
Fix for ["when multiple inputs are present and only one field is empty, all fields are highlighted in red, which is confusing"](https://linear.app/ducesnet/issue/FE-300/add-keywords-records-validation)

We could consider also adding this to the entire invenio ArrayField component as a wrapper around it to be used in all similar Form components, or are the Subjects an invidiual case?


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/MultilingualTextInput/MultilingualTextInput.jsx
```diff
@@ -41,6 +41,16 @@ export const MultilingualTextInput = ({
   const usedLanguages = usedSubValues(value);
 
   useShowEmptyValue(fieldPath, defaultNewValue, showEmptyValue);
+
+  useEffect(() => {
+    const fieldDOMNode = document.querySelector(
```
mirekys, line 46: Directly manipulating DOM from React should be avoided unless absolutely necessary

mirekys: Ref should be used instead

corovcam: The thing with Ref here is that Invenio doesn't forward it to the DOM element I need to target (the top Form.Field element of the ArrayField). 
Also for some reason Invenio passes empty `{}` object to the `error` prop of the SUI's `Form.Field` component, which makes the whole field red since an empty object is truthy. You can also see it in the console that it is a validation error, since the `Form.Field` requires a boolean value for `error` rather than an object.

mirekys: Querying whole document just by class from React code feels very unreliable, I would start at least by Ref'ing what we can (`ArrayField`), and then use the ref's.current to drill down and access its correct children (`Form.Field`). Agreed we could make some wrapper or atleast hook from this logic for ArrayFields. @Ducica what do you think?

corovcam: > Could you please also report this issue to react-invenio-forms repo? :) The way they set `{error: {}}` and set that to `Form.Field` also doesn't make sense to me

Yes, I will report this to them.

mirekys (changes requested):
For this to get merged, please address the issue with targeting the exact field element using ref of the parent (GH Copilot AI also mentions this). Class-based document query selector is not reliable for more complex forms...

mirekys (approved):

mirekys:
Thanks a lot! Looks good now as is :)

mirekys (approved):

PR #305, opened 2025-04-15
using doi from pids section

Ducica:


oarepo_ui/templates/components/IdentifiersAndLinks.jinja
```diff
@@ -1,9 +1,23 @@
-{# def originalRecordUrl=None, objectIdentifiers=None #}
-
-{% if objectIdentifiers|length > 0 or originalRecordUrl %}
+{# def originalRecordUrl=None, objectIdentifiers=None, repositoryAssignedIdentifier=None  #}
+{% if objectIdentifiers|length > 0 or originalRecordUrl or repositoryAssignedIdentifier %}
```
mirekys, line 2: accept pids and render all pids here, to make component more universal. For now we have only single DOI as managed PID, for other repos there may be more

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/components/RecordVersionsList.jsx
```diff
@@ -59,17 +72,14 @@ const RecordVersionItem = ({ item, activeVersion, searchLinkPrefix = "" }) => {
...
-                href={`https://doi.org/${doi}`}
-                className={
-                  "doi" + (activeVersion ? " text-muted-darken" : " text-muted")
-                }
-              >
-                {doi}
-              </a>
-            </div>
+            <a
+              href={`https://doi.org/${doi}`}
```
mirekys, line 76: https://github.com/oarepo/oarepo-ui/pull/305/files#diff-98d6ddf63d0febb2eb64373049344d2258c6bc0089d939df2c367f7c8da2b16aR8 from here I understood identifier could be either full URL or just the value? It should be handled consistently accross components

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/components/RecordVersionsList.jsx
```diff
@@ -6,16 +6,29 @@
...
 import { ErrorMessage } from "react-invenio-forms";
 
 import { httpApplicationJson } from "..";
 
+// unfortunately, Trans component and interpolation, have some weird issue, where
+// the / in the doi gets escaped, and even though you tell it to not escape, it still escapes it
+const ParentDoiMessage = ({ doi }) => (
+  <p className="text-muted">
+    <strong>{i18next.t("Cite all versions?")}</strong>{" "}
+    {i18next.t("You can cite all versions by using the DOI")} {doi}.{" "}
```
mirekys, line 22: Missing link to parent record doi, if we don't have it, don't mention citing by using it here

mirekys: There are two different messages being replaced with this component having only single message

oarepo_ui/templates/components/IdentifiersAndLinks.jinja
```diff
@@ -1,9 +1,23 @@
...
 <section aria-label='{{ _("Identifiers and links") }}' class="ui segment identifiers-and-links">
   <h2 class="ui small header detail-sidebar-header">{{_('Identifiers and links')}}</h2>
   <div class="ui link list">
+    {% if repositoryAssignedIdentifier %}
+    {% set _data = {
+      "href": repositoryAssignedIdentifier if repositoryAssignedIdentifier.startswith('https')
```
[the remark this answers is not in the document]
Ducica: will sanitize, can you explain the other part please?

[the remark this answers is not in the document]
mirekys: I think IdentifiersAndLinks as of a generic component, it should accept & render any identifier under pids field. For docs, there is DOI only, but for other repos, there may be multiple of different type

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/components/RecordVersionsList.jsx
Ducica: Ideally, whenever we get something that can be an url, I feel like URL should be constructed on BE and sent that way to front end, but unfortunately, that is not the case. I see what you mean, and I will treat it as everywhere in regards to identifier coming in pids, it is just the identifier and not the URL. Thanks!

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/components/RecordVersionsList.jsx
Ducica: My bad I apologize, I am blind

oarepo_ui/templates/components/IdentifiersAndLinks.jinja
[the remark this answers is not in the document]
mirekys: Current logic forces as https://doi.org url for any identifier value

Ducica:
@mirekys Ahoj, I think this one + https://github.com/Narodni-repozitar/nr-docs/pull/302/files are ready now. Please when you can, take a look

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/components/RecordVersionsList.jsx
```diff
@@ -6,15 +6,54 @@
...
+import { httpVnd } from "..";
+
+// unfortunately, Trans component and interpolation, have some weird issue, where
+// the / in the doi gets escaped, and even though you tell it to not escape, it still escapes it
+const ParentDoiMessage = ({ parentDoiObject }) => {
+  return (
+    <p className="text-muted">
+      <strong>{i18next.t("Cite all versions?")}</strong>{" "}
+      {i18next.t("You can cite all versions by using the DOI")}{" "}
+      {parentDoiObject?.url ? (
```
mirekys, line 24: move this one line above (text above should also be conditional)

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/components/RecordVersionsList.jsx
```diff
@@ -42,35 +82,42 @@ const RecordVersionItem = ({ item, activeVersion, searchLinkPrefix = "" }) => {
         <List.Header>
           {activeVersion ? (
             <span className="text-break">
-              {item.version_note ??
-                i18next.t("Version {{- version}}", {
-                  version: item.version_note || item.version,
-                })}
+              {/* As we now return a number in UI serialization for metadata.version,
```
mirekys, line 85: Even for version_note it is always a number?

mirekys (approved):

PR #312, opened 2025-04-25
Matomo Integration

xsenyaaax:
Use Matomo to track user activities on site.

All changes:
- Added new block to base age html
-  Added matomo_analytics.html with tracking code. If Matomo URL is defined, insert the code
- Changes in ext.py to include tracking template


oarepo_ui/templates/oarepo_ui/matomo_analytics.html
```diff
@@ -0,0 +1,14 @@
+{% if config.MATOMO_ANALYTICS_URL %}
+<script>
```
mirekys, line 2: This will break on production where in-line scripts like this are prohibited by CSP...should be bundled and refrerenced using webpack.py

xsenyaaax: > This will break on production where in-line scripts like this are prohibited by CSP...should be bundled and refrerenced using webpack.py

So this would not work even if I add to invenio.cfg:
```
APP_DEFAULT_SECURE_HEADERS["content_security_policy"]["default-src"].append(
    'https://127.0.0.1:8080',
)
```

With added headers it was sending GET requests locally on nr-docs and was not blocked by CSP

mesemus (approved):

mesemus (approved):

PR #321, opened 2025-06-05
added overrides for invitations page

Ducica:


oarepo_ui/ext.py
```diff
@@ -154,6 +154,13 @@ def _global_search_ui(app_name):
                     "CommunityInvitationsModal", "@js/communities_components"
                 )
             },
+            "oarepo_communities.community_invitations": {
+                "InvenioCommunities.CommunityMembers.InvitationsModal": UIComponent(
+                    "CommunityInvitationsModal",
+                    "@js/communities_components",
+                    props={"resetQueryOnSubmit": True},
```
mirekys, line 161: Why this extra prop not on regular modal for community members above?  I would share the same uicomponent for both cases if possible

PR #331, opened 2025-06-26
feature: sending additional facet labes to search config

Ducica:
*if someone clicks on search link and that facet is not in aggs in search config it will not be handler properly in UI

### DIFF
--- a/oarepo_ui/resources/config.py
+++ b/oarepo_ui/resources/config.py
@@ -4,6 +4,9 @@ from pathlib import Path
 import marshmallow as ma
 from flask import current_app
 from flask_resources import ResourceConfig
+from flask_resources import (
+    resource_requestctx,
+)
 from invenio_base.utils import obj_or_import_string
 from invenio_pidstore.errors import (
     PIDDeletedError,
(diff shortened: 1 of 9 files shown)


oarepo_ui/resources/config.py
```diff
@@ -185,6 +213,45 @@ def search_active_facets(self, api_config, identity):
...
 
+    def additional_filter_labels(self):
+        """
+        Returns human-readable list of filters that are currently applied in the URL.
+        Sometimes those are not available in the response from the search API.
+        """
+        translated_params = {}
+        facets = {}
+
+        for model in current_app.config.get("GLOBAL_SEARCH_MODELS", []):
```
mirekys, line 224: Replace with `current_global_search. global_search_model_services` from oarepo-global-search

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/ActiveFiltersElement.jsx
```diff
@@ -7,16 +7,24 @@ import { withState } from "react-searchkit";
...
 
-const getLabel = (filter, aggregations) => {
+const getLabel = (filter, aggregations, additionalFilterLabels) => {
   const aggName = filter[0];
   let value = filter[1];
+  // keep the original methodo of getting labels for backwards compatibility just in case
+  // to not break existing applications
   const label =
     aggregations[aggName]?.buckets?.find((b) => b.key === value)?.label ||
+    additionalFilterLabels[aggName]?.value_labels?.find((b) => b.key === value)
```
mirekys, line 17: Why not to store it also under `buckets`? Then this code could be extracted into `getValueLabel` helper to make things more readable

oarepo_ui/resources/config.py
Ducica: Hey, does oarepo-ui depend on oarepo-global-search?

PR #342, opened 2025-07-31
feat: display DOI on tombstone page

Ducica:
DOI
image
NO DOI
image

### DIFF
--- a/oarepo_ui/resources/resource.py
+++ b/oarepo_ui/resources/resource.py
@@ -38,6 +38,9 @@ from oarepo_runtime.datastreams.utils import get_file_service_for_record_class
 from oarepo_runtime.resources.responses import ExportableResponseHandler
 from werkzeug.exceptions import Forbidden
 
+from idutils import to_url
+
+
 from oarepo_ui.utils import dump_empty
 
 # Resource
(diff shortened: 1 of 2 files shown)


oarepo_ui/templates/oarepo_ui/Tombstone.jinja
```diff
@@ -22,10 +22,24 @@
     <table class="ui table">
       <tbody>
         {% for key, value in tombstone.items() %}
-          <tr>
-            <th class="left aligned pl-5">{{ _(key) }}:</th>
-            <td>{{ value }}</td>
-          </tr>
+        {% if key and value %}
+          {% if key == "URL"  %}
```
mirekys, line 26: move this to line 30, to share as much markup as possible

oarepo_ui/resources/resource.py
```diff
@@ -788,16 +791,28 @@ def tombstone(self, error, *args, **kwargs):
...
+            "self_html", None
+        )
 
         tombstone_dict = {}
         if record_tombstone:
             tombstone_dict = {
                 "Removal reason": record_tombstone["removal_reason"]["id"],
                 "Note": record_tombstone.get("note", ""),
                 "Citation text": record_tombstone["citation_text"],
+                "URL": tombstone_url,
```
mirekys, line 815: Should we render URL if it has None value (from logic above, that is possible)?

Ducica: We only render a row, if both key and value are truthy

mirekys: You're right, missed that, thanks :)

mirekys (approved):

PR #350, opened 2025-08-11
RecordActionButtons: merge ShareButton, EditButton and RecordRequests into single entrypoint

corovcam:


Ducica:
@corovcam is this still alive? if not, please close. Thanks

corovcam:
> @corovcam is this still alive? if not, please close. Thanks

To be honest, I am not really sure, but since we are trying to use all Invenio components now for new repos, I think it doesn't really make sense to have separate implementation for what they already have in the sidebar.

PR #352, opened 2025-08-13
Pylint

mesemus:


tests/conftest.py
```diff
@@ -180,15 +178,9 @@
...
-        user1 = datastore.create_user(
-            email="user1@example.org", password=hash_password("password"), active=True
-        )
-        user2 = datastore.create_user(
-            email="user2@example.org", password=hash_password("password"), active=True
-        )
-        admin = datastore.create_user(
-            email="admin@example.org", password=hash_password("password"), active=True
-        )
+        user1 = datastore.create_user(email="user1@example.org", password=hash_password("password"), active=True)
```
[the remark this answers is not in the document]
mirekys: Just for tests

oarepo_ui/templating/catalog.py
```diff
@@ -0,0 +1,459 @@
...
+        self.file_ext = file_ext
+        self.use_cache = use_cache
+        self.auto_reload = auto_reload
+
+        root_url = root_url.strip().rstrip(SLASH)
+        self.root_url = f"{root_url}{SLASH}"
+        env = flask.templating.Environment(undefined=jinja2.Undefined, app=current_app, autoescape=True)
+        extensions = [*(extensions or []), "jinja2.ext.do", JinjaX]
+        globals = globals or {}  # noqa: A001
+        current_app.config.setdefault(
```
mirekys, line 102: Shouldn't this belong somewhere else? Site-wide config.py?

oarepo_ui/templating/catalog.py
```diff
@@ -0,0 +1,459 @@
...
+class SearchPathItem(NamedTuple):
+    """Named tuple for search path items in the Jinjax catalog."""
+
+    template_name: str
+    absolute_path: Path
+    relative_path: Path
+    priority: int
+
+
+class OarepoCatalog(Catalog):
```
mirekys, line 57: Do we need to override so many things here? Many methods look very similar to: https://github.com/jpsca/jinjax/blob/main/src/jinjax/catalog.py

PR #363, opened 2025-09-16
fixed config variable name

Ducica:


oarepo_ui/views.py
```diff
@@ -54,6 +56,12 @@ def add_jinja_filters(state: BlueprintSetupState) -> None:
     return blueprint
 
 
+def create_rdm_templates_dummy_blueprint(app: Flask) -> Blueprint:  # noqa: ARG001
+    """Create a dummy blueprint for RDM templates."""
+    templates_path = Path(invenio_app_rdm.__file__).parent / "theme" / "templates"
+    return Blueprint("rdm_templates", __name__, template_folder=templates_path, static_folder="static")
+
```
Ducica, line 63: Please comment if this is OK approach for getting the templates path from invenio_app_rdm. I have tried using path such as invenio_app_rdm/theme/templates, but it does not seem to be able to pick up on that. Probably it somehow does the lookup relative to the file/folder where blueprint itself is being registered, so it needs absolute path?

PR #365, opened 2025-09-16
refactor modules & cleanup old code

mirekys:


oarepo_ui/config.py
```diff
@@ -22,12 +22,20 @@
 if TYPE_CHECKING:
     from collections.abc import Callable
 
-    from .ui.components import UIComponent, UIComponentOverride
+    from .overrides import UIComponent, UIComponentOverride
 
-# We set this to avoid https://github.com/inveniosoftware/invenio-administration/issues/180
-THEME_HEADER_LOGIN_TEMPLATE = "oarepo_ui/header_login.html"
+APP_THEME = ["semantic-ui"]
```
Ducica, line 27: Seems to be already here https://github.com/oarepo/oarepo/blob/rdm-13/oarepo/config/generic_parameters.py

oarepo_ui/config.py
```diff
@@ -22,12 +22,20 @@
...
+THEME_CSS_TEMPLATE = "oarepo_ui/css.html"
+INSTANCE_THEME_FILE = "./less/theme.less"
 
 OAREPO_UI_THEME_HEADER_FRONTPAGE = "oarepo_ui/header_frontpage.html"
+JAVASCRIPT_PACKAGES_MANAGER = "pnpm"
+ASSETS_BUILDER = "rspack"
+WEBPACKEXT_NPM_PKG_CLS = "pynpm:PNPMPackage"
+# TODO: update this when oarepo-global-search gets migrated to RDM13
+SEARCH_UI_SEARCH_TEMPLATE = "oarepo_ui/search.html"
+
```
Ducica, line 38: Probably similar as above could be put in oarepo which is then used in invenio.cfg in app copier https://github.com/oarepo/nrp-app-copier/blob/rdm-13/template/invenio.cfg.copier

mirekys: Thanks for pointing that out. When looking to generic parameters, it makes real good sense to put it there

Ducica (dismissed):

Ducica (approved):

PR #367, opened 2025-09-18
feat(templates): register all rdm template blueprints

mirekys:


oarepo_ui/rdm_templates.py
```diff
@@ -0,0 +1,125 @@
...
+    blueprint.add_app_template_filter(order_entries)
+    blueprint.add_app_template_filter(get_scheme_label)
+    blueprint.add_app_template_filter(has_images)
+    blueprint.add_app_template_filter(localize_number)
+    blueprint.add_app_template_filter(compact_number)
+    blueprint.add_app_template_filter(truncate_number)
+    blueprint.add_app_template_filter(namespace_url)
+    blueprint.add_app_template_filter(custom_fields_search)
+    blueprint.add_app_template_filter(transform_record)
+
```
Ducica, line 64: Thx for the PR. I guess these were mandatory to add so that templates would work? I think some of these will not work as intended for our case, especially the permissions related items. Is the lack of these causing the app to fully crash at blueprint registration time? If not, maybe it would make sense to not have these and then when we need to use this template, it will probably crash and then we can check if their implementation makes sense for us and uncomment potentially or replace with our own. If it crashes without these, then I think just leave it as it is, but we should be careful. @mesemus What do you think pls?

Ducica (approved):

PR #369, opened 2025-09-23
Add default fallback parameter to FieldData label, help, and hint

xsenyaaax:


mesemus:
What is the way of calling these staticmethods from jinja templates? I can't seee them registered in global filters, maybe just overlooked.

PR #376, opened 2025-10-07
Testing form wizard

Ducica:
First version of the form wizard. 

Note, that code contains some inline styling. I left it like this, because we need to make order in less components (as we have less components in oarepo-ui, that are on the "default" path, but these are no longer used by theme.config file and I thought it would be too many changes for one pull request. Plus we need to decide if we will always use rdm forlder name, or if we will also look for styles in default path. 

To test the form you can use datarepo https://github.com/NRP-CZ/datarepo/tree/testing-requests

Issues not resolved in the PR:

Where the actions will be and what actions will be displayed - there are a few unknowns in this regard (requests/norequests - or just a few form actions).
Mobile view (in my opinion, the tabs will have to switch to horizontal layout and be above the content panel in that case, which would mean you would get a horizontal scroll on tabs.

This will break existing repositories that are on v14.


mirekys (dismissed):

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/BaseFormLayout/BaseFormLayout.jsx
```diff
@@ -40,36 +41,47 @@ export const FormTitle = () => {
...
       window.removeEventListener("beforeunload", handleBeforeUnload);
     };
-  }, []);
+  }, [dirty]);
 
-  return (
+  return useWizardForm ? (
+    <Grid>
+      <Grid.Column id="main-content" mobile={16} tablet={16} computer={16}>
+        {/* TODO: do we really need to display the title? I think this is a redundant information */}
```
mirekys: I agree with you

mirekys (dismissed):

Alzpeta (approved):

PR #421, opened 2026-03-13
passing embedded tag to detail page

Ducica:


oarepo_ui/resources/records/resource.py
```diff
@@ -370,7 +371,7 @@ def _detail(
         return response
 
     @pass_route_args("view")
-    @pass_query_args("record_detail")
+    @pass_query_args("record_detail", "embed")
```
mirekys: Should be part of record_detail args, like `preview`, etc. : https://github.com/oarepo/oarepo-ui/blob/b57c8554297b8151fbbce397270c12bfcf1333c1/oarepo_ui/resources/records/config.py#L146

mirekys (changes requested):

mirekys (dismissed):

oarepo_ui/resources/records/config.py
mirekys: What happens to this? It now seems redundant, unused? Or is this used anywhere?

mirekys: If not, should be removed

oarepo_ui/templates/oarepo_ui/pages/RecordDetail.jinja
```diff
@@ -18,5 +18,6 @@
   user_avatar = None,
   record_owner_id = None,
   user_communities_memberships = {},
+  embedded = False,
 #}
 {% extends model_name ~ "/record_detail.html" %}
```
mirekys: nit: missing newline

oarepo_ui/resources/records/config.py
Ducica: Seems it is a leftover from the past. Thanks

mirekys (approved):

PR #428, opened 2026-04-01
removed workflow/requests dependency

Ducica:


tests/test_deposit_create.py
```diff
@@ -69,7 +69,7 @@ def test_deposit_create_without_community(app, logged_client, users, extra_entry
         assert response["preselectedCommunity"] is None
 
 
-def test_deposit_create_with_community(app, logged_client, users, extra_entry_points, init_communities_cf, community):
+def test_deposit_create_with_community(app, logged_client, users, extra_entry_points, community):
```
Ducica, line 72: as I understood, the basic fixture community should create a community without wf/cf now in latest pytest-oarepo

mesemus (approved):

PR #438, opened 2026-04-16
section fillness indicator

Ducica:
Originally, when we discussed, we agreed to use the check mark icon to be orange when section is not filled. However, as we have error indication there and it depends on the Backend and as the fillness indicator should be purely FE related (otherwise, if you have an error, you would not see fillness change while filling the section before you save), instead I put a thin progress bar under each tab/step. I think it works fine. Also, I decided not to use semantic ui progress bar, because it has weird behavior where it is a little bit filled, even when you sat that it is 0% full, which I found really weird. I will fix up the comments from copilot, but for now, I would like to just confirm if this approach is acceptable or not. 

https://github.com/NRP-CZ/ccmm-invenio/pull/28


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/SectionCompletionBar/SectionCompletionBar.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+const DEFAULT_FILLED_THRESHOLD = 0.8;
+
+export const SectionFilledBar = ({
+  includesPaths,
+  sectionFilled,
+  filledThreshold = DEFAULT_FILLED_THRESHOLD,
+}) => {
+  const { values } = useFormikContext();
+  const reduxState = useSelector((state) => state);
```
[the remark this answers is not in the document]
mirekys: Yes.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/TabForm/TabForm.jsx
```diff
@@ -36,13 +36,15 @@ export const TabForm = ({ sections = [] }) => {
   const [activeStep, setActiveStepState] = React.useState(
     Math.max(initialStep, 0)
   );
-  const [isTransitioning, setIsTransitioning] = React.useState(false);
-  const transitionTimerRef = useRef(null);
+  const [contentVisible, setContentVisible] = React.useState(true);
+  const [isSwapping, setIsSwapping] = React.useState(false);
+  const pendingStepRef = useRef(null);
```
mirekys: Unused, right?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/TabForm/TabForm.jsx
```diff
@@ -65,19 +66,22 @@ export const TabForm = ({ sections = [] }) => {
...
-      setIsTransitioning(true);
-      if (transitionTimerRef.current) {
-        clearTimeout(transitionTimerRef.current);
-      }
-      transitionTimerRef.current = setTimeout(() => {
-        transitionTimerRef.current = null;
-        setActiveStepState(index);
-        setIsTransitioning(false);
-      }, 0);
+      setActiveStepState(index);
```
mirekys: Race condition on rapid tab switching, `handleSetStep` sets `contentVisible(false)` but does not clear the pending swap timer from a previous transition. If a user clicks tabs quickly, the stale `setTimeout` from `handleTransitionHide` fires and incorrectly sets `isSwapping(false)` + `setContentVisible(true)`, causing content to appear mid-fade or in an inconsistent state. The previous implementation cleared `transitionTimerRef.current` on each step change; this refactor dropped it.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/util.js
```diff
@@ -402,3 +403,74 @@ export function categorizeErrors(errors) {
...
+ *
+ * @param {Object} params
+ * @param {Object} params.formikValues - Formik values object
+ * @param {Object} params.reduxState - Redux store state
+ * @param {string[]} params.includesPaths - Array of field paths to check
+ * @returns {number} Fraction filled (0 to 1)
+ */
+export const computeSectionCompletion = ({
+  formikValues,
+  reduxState,
```
mirekys: reduxState ignored in implementation?

Ducica: Yes the generic implementation only looks in formik, but generally for some sections you would need to look into redux store, so I thought to make the interface consistent between generic and some more specific callback you might add for a section, I can remove if you think it does not make sense.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/TabForm/TabForm.jsx
Ducica: Thanks!

mirekys:
Pls rebase

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/SectionCompletionBar/SectionCompletionBar.jsx
```diff
@@ -0,0 +1,58 @@
...
+  const percentage = Math.round(filledRatio * 100);
+  const color = filledRatio < sectionCompletionThreshold ? "orange" : "green";
+  const filledText = i18next.t("{{percentage}}% of this section is filled", {
+    percentage,
+  });
+  if (percentage === 0) {
+    return null;
+  }
+  return (
+    <div
```
mirekys, line 35: Let's leave this tip for later, actually a valid point by sonar

oarepo_ui/translations/en/LC_MESSAGES/messages.po
```diff
@@ -1481,3 +1481,9 @@ msgstr ""
 msgid ""
 "Section definition is missing the component key and it must be defined."
 msgstr ""
+
+msgid "{{percentage}}% of this section is filled"
```
mirekys, line 1485: This keys shouldn't be needed anymore?

oarepo_ui/translations/cs/LC_MESSAGES/messages.po
```diff
@@ -1510,3 +1518,9 @@ msgstr "Tato sekce není definována."
 msgid ""
 "Section definition is missing the component key and it must be defined."
 msgstr "V definici sekce chybí klíč component, který musí být definován."
+
+msgid "{{percentage}}% of this section is filled"
+msgstr "{{percentage}}% této sekce je vyplněno"
+
+msgid "Section completion"
+msgstr "Dokončení sekce"
```
mirekys: Weird cs translation?

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/TabForm/TabForm.jsx
```diff
@@ -56,6 +48,9 @@ export const TabForm = ({ sections = [] }) => {
       if (!(index >= 0 && index < sectionKeys.length)) {
         return;
       }
+      if (index === activeStep && pendingStep === null) {
```
mirekys, line 55: confusing naming, what does pending step is it waiting for something? is it actually next step to be taken?

oarepo_ui/translations/en/LC_MESSAGES/messages.po
Ducica: It is still used as aria for the progress bar

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/TabForm/TabForm.jsx
Ducica: Pending step is the step you are switching to. The flow is setPending step to the step you clicked on, setContentVisible to false, to smoothly unmount the content, then transition to the new step. To be honest, all of this is to avoid the bug with html5 backend, because invenio uses dnd providers on components instead of the entire app. This was already solved, but the transition was kind of janky, so I thought to add this to make it smoother. I did not find any issues while testing.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/SectionCompletionBar/SectionCompletionBar.jsx
Ducica: If I understood correctly, you mean to use <progress> instead of div? I've already tried it and you need to do things like                                                                                                               
    &::-webkit-progress-bar { background-color: @sectionCompletionBarBackgroundColor; }                          
    &.green::-webkit-progress-value { background-color: @sectionCompletionBarFilledColor; }
    &.orange::-webkit-progress-value { background-color: @sectionCompletionBarWarningColor; }                    
    &.green::-moz-progress-bar { background-color: @sectionCompletionBarFilledColor; }                           
    &.orange::-moz-progress-bar { background-color: @sectionCompletionBarWarningColor; }  

for it to work normally. I'd rather not go this way tbh, but please let me know

mirekys: That looks like a normal progress element styling, I think it's more semantic html friendly than just plain div. If you already had it styled to look & behave correctly, I'd prefer progress element, if not, ignore for now :)

oarepo_ui/translations/en/LC_MESSAGES/messages.po
mirekys: You're right, I missed that 🙈

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/TabForm/TabForm.jsx
mirekys: Thanks for explanation, could you please add the comment here into code, explaining why this pending step mechanism was needed?

oarepo_ui/translations/cs/LC_MESSAGES/messages.po
```diff
@@ -1510,3 +1518,18 @@ msgstr "Tato sekce není definována."
 msgid ""
 "Section definition is missing the component key and it must be defined."
 msgstr "V definici sekce chybí klíč component, který musí být definován."
+
+msgid "{{percentage}}% of this section is filled"
+msgstr "{{percentage}}% této sekce je vyplněno"
+
+msgid "Section completion"
+msgstr "Vyplnění sekce"
```
mirekys, line 1526: Vyplněnost sekce

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/TabForm/TabForm.jsx
Ducica: Done

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/SectionCompletionBar/SectionCompletionBar.jsx
```diff
@@ -0,0 +1,58 @@
...
+        });
+
+  const percentage = Math.round(filledRatio * 100);
+  const color = filledRatio < sectionCompletionThreshold ? "orange" : "green";
+  const filledText = i18next.t("{{percentage}}% of this section is filled", {
+    percentage,
+  });
+  if (percentage === 0) {
+    return null;
+  }
```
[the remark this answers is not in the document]
Ducica: Agrreed explicitly to do it this way

mirekys (approved):

PR #439, opened 2026-04-16
passing form title

Ducica:


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/TabForm/TabForm.jsx
mirekys, line 177: How does this look on smaller screens? Form title should also be displayed there somewhere?

Ducica: Smaller screen is not in the design and honestly I don't see a good place to put it. I was thinking it does not necessarily have to be there for smaller screens. If yes, we would have to put it on top left I think right above steps, but it looks kind of weird.

mirekys (approved):

PR #441, opened 2026-04-20
fix: pages prefix duplicated

mesemus:


mirekys:
Where did duplication occur?

Alzpeta (approved):

PR #446, opened 2026-04-24
fix: make buttons stack on top one another on mobile size

Ducica:


mirekys:
I'd rather see this implemented using SUI components than custom less hacks - simply wrap the buttons in a stackable grid: https://react.semantic-ui.com/collections/grid/#responsive-variations-stackable and that should do the trick if I'm not overlooking something

mirekys (approved):

PR #460, opened 2026-05-20
[breaking] fixed default search app layout

Ducica:
Made default search app layout like we usually get in a design:

Namely, search bar in the content, in the right hand column. Otherwise, when using invenio's layout (search bar in header), it remains the same. Also replaced some existing default components by those that they almost always do a certain way in the design. Put "Go to top" button as a default component in the layout. If someone does not want it they can just disable it via css.  Note that button is not inside SearchAppLayout, because it has a state setter, and their react portal that they use, gets really buggy (for the scenario where searchbar is teleported to header). Deleted some components we are not using anywhere anymore.

The goal:

We will say we support search bar in header or search bar in content in right hand column -> other layouts we dont do. 

https://github.com/NRP-CZ/datarepo/pull/56
Old style layout:
image

New style layout 
image


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/ActiveFilters.jsx
```diff
@@ -0,0 +1,103 @@
...
+  const aggName = filter[0];
+  const value = filter[1];
+
+  const aggLabel =
+    aggregations[aggName]?.label ||
+    additionalFilterLabels[aggName]?.label ||
+    aggName;
+
+  const _getValueLabel = (aggs) =>
+    aggs[aggName]?.buckets?.find((b) => b.key === value)?.label;
```
corovcam: To make it more type-safe
```suggestion
    aggs?.[aggName]?.buckets?.find((b) => b.key === value)?.label;
```

corovcam (commented):
So basically you just coppied an already functioning design from repos (datarepo) that we use often and added it here with minor changes, as I understand correctly? Otherwise looks good.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppFacets.jsx
```diff
@@ -39,30 +39,42 @@ SearchAppFacetsFallback.propTypes = {
...
+          {aggs.map((agg) => (
+            <ErrorBoundary
+              FallbackComponent={(props) => (
+                <SearchAppFacetsFallback {...props} agg={agg} />
+              )}
+              key={agg.aggName}
+            >
+              <BucketAggregation
+                key={agg.aggName}
+                title={agg.titles}
```
corovcam: agg.titles or agg.title? because https://github.com/oarepo/oarepo-ui/blob/1dc6fe2a53d268e13138e9ddac0b75591ce51972/oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppFacets.jsx#L16-L17 uses agg.title

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/ActiveFilters.jsx
Ducica: Thanks. Yes I took what active filters component looks like in design for datarepo and made it default active filters component. And this is the one we will use from now on, so if anyone wants to deviate from this, then they need to make their own component.

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/search/SearchAppFacets.jsx
Ducica: Thanks, you are right

PR #463, opened 2026-05-27
[breaking] fix: unable to create urls for home page

Ducica:


mesemus (dismissed):
There is a lot of copied code from over the RDM in this file. We should consider either:

1. importing stuff that we can (robots etc) from rdm and reuse it
2. patch invenio-app-rdm to allow registering our own view functions and register those via config variables

Is there a reason for having copied so much code or can we implement 1. in this PR?

Ducica:
> There is a lot of copied code from over the RDM in this file. We should consider either:
> 
> 1. importing stuff that we can (robots etc) from rdm and reuse it
> 2. patch invenio-app-rdm to allow registering our own view functions and register those via config variables
> 
> Is there a reason for having copied so much code or can we implement 1. in this PR?

TBH we had this discussion in the past, but I don't remember exactly why we did it this way. @mirekys  Do you remember by some chance please?

mesemus (dismissed):

mesemus (approved):

mesemus (approved):

mesemus (dismissed):

PR #465, opened 2026-06-02
fix: fixed form inside modal styling

Ducica:


oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/TabContent/TabContent.jsx
```diff
@@ -128,6 +120,14 @@ export const TabContent = ({ activeStep, sections, next, back }) => {
             initialRecord={initialRecord}
           />
         </div>
+        {dirty && (
+          <Message info data-testid="unsaved-changes-message">
```
Ducica, line 124: Moved the information about unsaved changes to the bottom. Original motivation was that, when you have a field with drag and drop right at the top of the section, dragging touches formik state and so the message appears and can cause buggy behavior of dnd because layout shifts. I think it even makes more sense to have it down closer to  action buttons.

oarepo_ui/theme/assets/semantic-ui/less/oarepo_ui/themes/rdm/views/item.overrides
```diff
@@ -137,3 +137,13 @@
     }
   }
 }
+.ui.modal {
+  .item.deposit-drag-listitem {
+    .content {
+      width: auto !important;
```
Ducica, line 143: Some invenio components are not really used inside of modals and when you put them in a modal, certain styles clash with SUI modal styling, like for example creatibutor field items that use .content class, but modal also uses some styling for .content (the edit/delete buttons next to each list item, get misplaced).

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/EDTFDatePickerField/EDTFDateRangePickerField.jsx
```diff
@@ -139,11 +140,12 @@ export const EDTFDaterangePicker = ({
           setDateEdtfFormat={setDateEdtfFormat}
           dateFormat={dateFormat}
           datePickerProps={{ ...pickerProps, ...datePickerPropsOverrides }}
+          customInputProps={{
+            helpText: fieldData.helpText,
+            ...customInputProps,
+          }}
```
Ducica, line 146: Made slight changes to date picker, so that the structure of dom would be the same as for invenio components. Otherwise, you need to do a lot of battling with styling we are inheriting from rdm/invenio theme

oarepo_ui/theme/assets/semantic-ui/js/oarepo_ui/forms/components/TabContent/TabContent.jsx
mirekys: I think best way would be to have some modal pop-up fixed on top of viewport? Not causing any layout shifts?

mirekys: But this solution fine for now

mirekys (approved):

PR #469, opened 2026-06-10
fix: fixed returning array with empty sentinel when no data

Ducica:


mirekys (dismissed):

oarepo_ui/templating/data.py
mirekys, line 391: might add the samne sentinel check here aswell?

Ducica: Done thanks

mirekys (approved):

mesemus (approved):

PR #473, opened 2026-06-17
feat: preselecting community if user is member of only one

Ducica:


oarepo_ui/resources/decorators/pass_draft_community.py
```diff
@@ -38,6 +41,11 @@ def view(self: RecordsUIResource, **kwargs: Any) -> Any:
         comid = request.args.get("community")
         community = None
 
+        if not comid:
+            memberships = get_user_communities_memberships()
```
Ducica, line 45: Is it safe to do, are we going to have users with a lot of communities?

mesemus: Maybe in nr-docs, in most of the repositories it is safe, so I'd keep it for now and optimize when needed.

mesemus (approved):

PR #476, opened 2026-06-18
BE-1102: remove delete_draft from deposit create permissions

mirekys:
## Summary

Upstream fix for BE-1102: the new-record deposit page (/uploads/new) no longer advertises the `delete_draft` permission, because there is no record to delete yet. The Delete button remains available on the edit deposit page for real drafts.

## Changes

- `oarepo_ui/resources/records/config.py`: removed `delete_draft` from `deposit_create_permissions` (kept in `deposit_edit_permissions`).
- `tests/test_deposit_create.py`: updated the create-page permission expectation to no longer include `can_delete_draft`.

## Verification

- `./run.sh lint` passed (ruff, mypy, pyright).
- `./run.sh jslint` passed (`TabForm.jsx` is clean; one pre-existing warning in `ActiveFilters.jsx`).
- `TabForm.jsx` already guards `DeleteButton` with `permissions?.can_delete_draft && record?.id` — no change needed.
- `resource.py` already passes `dashboard_routes` to form config in both `_create()` and `_edit()` — no change needed.

### DIFF
--- a/oarepo_ui/resources/records/config.py
+++ b/oarepo_ui/resources/records/config.py
@@ -238,7 +238,6 @@ class RecordsUIResourceConfig(UIResourceConfig):
     deposit_create_permissions: ClassVar[list[str]] = [
         "manage",
         "manage_files",
-        "delete_draft",
         "manage_record_access",
     ]
     """List of permission actions to check for deposit create page."""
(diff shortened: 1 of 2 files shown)


Ducica:
Is the correct solution not to just make the permission work on /new and then the user can delete draft right after they save it without needing a full page refresh?

PR #480, opened 2026-06-23
unified permissions

Ducica:
We had permissions being checked in different ways. Something was in config variables, something was in ui resource/config. IMHO, also no reason to differentiate between drafts/records permissions because appropriate service handles those and must return correct values. Made a unification, so that permissions would behave the same. Especially important for the form, where it was rather confusing.

### DIFF
--- a/oarepo_ui/resources/records/config.py
+++ b/oarepo_ui/resources/records/config.py
@@ -11,7 +11,7 @@
 from __future__ import annotations
 
 import inspect
-from typing import TYPE_CHECKING, Any, ClassVar, NotRequired, TypedDict, cast
+from typing import TYPE_CHECKING, Any, NotRequired, TypedDict, cast
 
 import marshmallow as ma
 from flask import current_app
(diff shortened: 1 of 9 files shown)


oarepo_ui/resources/records/config.py
```diff
@@ -213,36 +213,6 @@ class RecordsUIResourceConfig(UIResourceConfig):
     """Field data item getter for retrieving field data items in the UI.
     If not set, the default getter will be used."""
 
-    record_detail_permissions: ClassVar[list[str]] = [
```
mirekys, line 216: Breaking change for anyone extending this base class? Or these were not really used before this PR?

oarepo_ui/resources/records/resource.py
```diff
@@ -269,7 +269,7 @@ def _detail(  # noqa: PLR0913 too many arguments
             "files": files_dict,
             "media_files": media_files_dict,
             "user_communities_memberships": get_user_communities_memberships(),
-            "permissions": (record.has_permissions_to(self.config.record_detail_permissions) if record else {}),
+            # "permissions" is populated by PermissionsComponent.before_ui_detail
```
mirekys: Remove comment

oarepo_ui/resources/records/config.py
mirekys: How we make sure permission component gets registered for anyone using this class ? so they suddenly don't loose all permissions when they forget to explicitly registeer Permissions comp

Ducica: They were being used, that is why we ended up having different sets of permissions in different locations i.e. later when you look at permissions on detail or form. If someone was extending the base class and reassigning value to this variable, then their variable would do effectivelly nothing, but they would still get permissions in the same spot as before, only they would contain more permissions - imho it is not a problem. In model copier, we are defining component sets for each model, the weakest type of model is empty model which contains

    AllowedHtmlTagsComponent,
        FilesEnabledComponent,
        BabelComponent,
        PermissionsComponent,
        FilesComponent,
        
        all the other models contain this + some more components. Unless someone would define their own components variable and not include this, I think no way to ommit this component.

mirekys (approved):

PR #494, opened 2026-07-28
Base filesLocked on the draft bucket lock (not is_published)

Ducica:
## What
`FilesLockedComponent.before_ui_edit` computed:

```python
form_config["filesLocked"] = not can_update_files or record.get("is_published", False)
```

For a published record `is_published` is always `True`, so the deposit form's uploader stayed **locked forever** — even after invenio's file-modification grace-period request unlocked the file bucket. Unlocking had no visible effect.

## Fix
Compute `filesLocked` from the draft's **actual bucket lock state** instead of `is_published`:

```python
draft = record_from_result(api_record)
bucket = getattr(getattr(draft, "files", None), "bucket", None)
form_config["filesLocked"] = (not can_update_files) or bool(getattr(bucket, "locked", False))
```

So: bucket locked → `filesLocked=True` (locked UI); bucket unlocked by the file-modification request → `filesLocked=False` (uploader editable on reload). Behaviour is unchanged for new drafts (unlocked bucket → editable) and still-locked published records.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/oarepo_ui/resources/components/files_locked.py
+++ b/oarepo_ui/resources/components/files_locked.py
@@ -26,52 +26,54 @@ if TYPE_CHECKING:
 
 
 class FilesLockedComponent[T: RecordsUIResourceConfig = RecordsUIResourceConfig](UIResourceComponent[T]):
-    """Add files locked to form config, to be able to use the same logic as in RDM."""
+    """Drive the ``files_locked`` template value.
+
+    Corresponds to the ``deposits-record-locked-files`` hidden input, to be able
+    to use the same single-source logic as in RDM.
+
+    The value is exposed only through the hidden input (read by ``parseFormAppConfig``
+    as the top-level ``filesLocked`` prop), matching upstream RDM -- it is deliberately
+    NOT duplicated into ``form_config`` (``deposits-config``), so the frontend has a
+    single, unambiguous source of truth.
+    """
 
     @override
     def before_ui_create(
         self,
         *,
-        form_config: dict,
+        render_kwargs: dict,
         **kwargs: Any,
     ) -> None:
-        """Set filesLocked to False before rendering the create page.
+        """Files are never locked on the create page.
 
-        :param record: UI-serialized record dictionary (unused for create).
-        :param data: Empty API-serialized record data for the create form.
-        :param identity: Current user identity.
-        :param form_config: Form configuration dictionary to mutate in-place.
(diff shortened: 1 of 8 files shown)


oarepo_ui/resources/components/files_locked.py
```diff
@@ -47,23 +52,26 @@ def before_ui_create(
...
         :param extra_context: Extra context expected to contain permissions.
         """
-        form_config["filesLocked"] = not extra_context.get("permissions", {}).get(
-            "can_update_files", False
-        ) or record.get("is_published", False)
+        can_update_files = extra_context.get("permissions", {}).get("can_update_files", False)
+        draft = record_from_result(api_record)
+        bucket = getattr(getattr(draft, "files", None), "bucket", None)
+        bucket_locked = bool(getattr(bucket, "locked", False))
+        form_config["filesLocked"] = not can_update_files or bucket_locked
```
mirekys: I'd rather set it to extra_context and align with RDM (with reasoning as per the file modification PR): https://github.com/inveniosoftware/invenio-app-rdm/blob/700995a9ad2f5a7a02694d2f86d028eedc758bcd/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/index.js#L25

mirekys: https://github.com/inveniosoftware/invenio-app-rdm/blob/700995a9ad2f5a7a02694d2f86d028eedc758bcd/CHANGES.rst?plain=1#L306

Ducica: Understood thanks

oarepo_ui/templates/oarepo_ui/deposit/form.html
```diff
@@ -39,6 +39,11 @@
           value='{{ config.RECORDS_RESOURCES_ALLOW_EMPTY_FILES | tojson }}'>
   <input type="hidden" name="deposits-use-uppy-ui"
           value='{{ config.APP_RDM_DEPOSIT_NG_FILES_UI_ENABLED | tojson }}'>
+  {# File-modification (grace-period) evaluation, kept as its own hidden input to
```
mirekys, line 42: Comment (probably claude :)) not needed here?

mirekys (approved):

